PR #25, opened 2017-02-15
Matlab code-base

gmrukwa:
This is MATLAB code-base for the project.

As almost none of us understand MATLAB swiftly, only the .gitignore and .NET naming convention should be checked here (in the .prj file of the compiled library) for exported functions and classes.

Compiled version of this project is being uploaded onto One Drive.


gmrukwa:
[MATLAB code-base](https://trello.com/c/O9bCuqCl/33-matlab-code-base)

.gitignore
```diff
@@ -1,3 +1,12 @@
+## Ignore MATLAB artifacts
+*.asv
+
+## Ignore MATLAB data files
+*.mat
+
+## Ignore MATLAB compiler output
+*-compile/
+
```
FreexD: https://www.gitignore.io/api/matlab wouldn't we need some of these as well?

MATLAB/algorithms.prj
```diff
@@ -0,0 +1,211 @@
+<deployment-project plugin="plugin.ezdeploy" plugin-version="1.0">
+  <configuration file="C:\Users\Grzegorz\Source\Repos\MSI-dev\msi-divik\MATLAB\algorithms.prj" location="C:\Users\Grzegorz\Source\Repos\MSI-dev\msi-divik\MATLAB" name="algorithms" target="target.ezdeploy.library" target-name="Library Compiler">
```
FreexD: The rest of the project members would't be needing that one, right? It's only for compilation on your machine? It contains an absolute path which concerned me a little.

.gitignore
gmrukwa: We won't compile MEX, nor use SIMULINK. I think those are not necessary. Possibly this one:
```
# OSX / *nix default autosave extension
*.m~
```
may be needed.

MATLAB/algorithms.prj
gmrukwa: This should be generic. I was convinced MATLAB supports any portability by default.

FreexD:
LGTM

Issue #9, opened 2017-02-09
Create .NET and JS solution


gmrukwa:

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Create .NET solution](https://trello.com/c/3v7cDsnq/27-create-net-solution)

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Create JS solution](https://trello.com/c/ZhEHlf0P/28-create-js-solution)

gmrukwa:
JS solution remains open topic.

.NET solution part closed as ddcbcf1.

gmrukwa:
Closed as 3008361f42c4e6eeca1a6f67c0702eec7a200580

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Push JS solution in current form](https://trello.com/c/l9y45wad/61-push-js-solution-in-current-form)

PR #29, opened 2017-02-17
Initial solutions

gmrukwa:
@FreexD created initial solution for .NET. It contains a WebApi project initialized in VS, and some DLL's we will be populating later in the project.

This PR has been recreated in order to re-trigger CI.

Partially solves #9


Spectre.Tests/Controllers/HomeControllerTest.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Web.Mvc;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Spectre;
+using Spectre.Controllers;
+
+namespace Spectre.Tests.Controllers
+{
+    [TestClass]
+    public class HomeControllerTest
```
Daroslaw: So... MSTest for C# confirmed?

gmrukwa: I think we may stick to that. We had no problems with it in previous projects.

PR #30, opened 2017-02-17
Readme improvements to master

Sand3r-:
Since this is the first thing that will be read by a newcomer, I felt like the readme should be slightly improved language-wise.


Sand3r-:
Many repositories use files such as AUTHORS or CONTRIBUTORS held in main folder where they store contributors using pattern of
Name Surname <email@example.com>
Name SecondName Surname <example@email.com>

Names are stored in alphabetical order, sorted by name (for some reason).

I think we can do that as well.

gmrukwa:
This PR should be the same as #32.

Sand3r-:
On the other hand, we can just redirect to [contributions](https://github.com/spectre-team/spectre/graphs/contributors), just like [BLVC/caffe](https://github.com/BVLC/caffe/blob/master/CONTRIBUTORS.md) does.

Advantages:
- doesn't need to be manually updated
- includes actual contribution impact

Disadvantages:
- some people may hide their real name from their github account, so they won't be given full credits until they reveal it.

I'd opt for this one for the sake of simplicity.

PR #68, opened 2017-03-26
Feature/divik result

### LINKED ISSUE #51: Implement DivikResult
The task is to implement DivikResult class corresponding 1:1 with its MATLAB counterpart (if possible).

@Sand3r-  - you are the man to consult MATLAB interfaces with.
@Sebaxp95  - in case of any trouble, feel free to ask

Regis128:
Resolves #51


gmrukwa:
[Implement DivikResult](https://trello.com/c/Su3f7zsc/71-implement-divikresult)

src/Spectre.Algorithms/Methods/Segmentation.cs
```diff
@@ -66,7 +66,7 @@ public DivikResult Divik(IDataset dataset, DivikOptions options)
 
 			var varargin = options.ToVarargin();
 			var tmp = _segmentation.divik(numberOfOutputArgs, dataset.GetRawIntensities(), coordinates, varargin);
-			var result = new DivikResult(tmp);
+			var result = new DivikResult(tmp[1]);
```
Sand3r-: I think some comment explaining why is it tmp[1] would be appreciated.

Sand3r-: Or preferably change the name to more self-explanatory.

src/Spectre.Algorithms/Results/DivikResult.cs
```diff
@@ -27,17 +28,197 @@ namespace Spectre.Algorithms.Results
 	/// </summary>
 	public sealed class DivikResult
     {
-	    private object _matlabResultStruct;
+        #region Properties
+        /// <summary>
+        /// clustering quality index
```
Sand3r-: Please start documentation summaries with an uppercase letter. (Here and in other properties below)

src/Spectre.Algorithms/Methods/Segmentation.cs
Regis128: tmp[1] because tmp[0] is the same as field Partition in tmp[1], so we don't need tmp[0] at all

gmrukwa: Please place the same explanation in code (no one in future will check this pull request) ;)

Sand3r-: And also name tmp could be changed to matlabResult, so that its name truely resemblance what it really is.

PR #69, opened 2017-03-26
JSON serialization for DivikResult implemented

Sand3r-:
JSON serialization for DivikResult implemented.

The tests for checking whether the file was saved properly have been added as well.

### DIFF
--- a/src/Spectre.Algorithms.Tests/Results/DivikResultTests.cs
+++ b/src/Spectre.Algorithms.Tests/Results/DivikResultTests.cs
@@ -38,7 +38,7 @@ namespace Spectre.Algorithms.Tests.Results
         public void SetUpFixture()
         {
             var dataset = new BasicTextDataset(_testFilePath);
-            var options = DivikOptions.ForLevels(1);
+            var options = DivikOptions.ForLevels(2);
             options.MaxK = 2;
             options.Caching = false;
             options.PlottingPartitions = false;
@@ -72,6 +72,17 @@ namespace Spectre.Algorithms.Tests.Results
             Assert.AreEqual(_result.VarianceFilter, deserialisedResult.VarianceFilter, "The VarianceFilter differs from original");
             Assert.AreEqual(_result.VarianceThreshold, deserialisedResult.VarianceThreshold, "The VarianceThreshold differs from original");
             Assert.AreEqual(_result.Subregions, deserialisedResult.Subregions, "The Subregions differ from original");
+            File.Delete(path);
+        }
+
+        [Test]
+        public void SavedIdented()
+        {
+            string path = TestContext.CurrentContext.TestDirectory + "\\..\\..\\..\\test-path.json";
+            _result.Save(path, false);
+
+            string contents = File.ReadAllText(path);
+            Assert.False(contents.Contains("\n"), "Non-idented formatting contains new line");
             File.Delete(path);
         }
     }
--- a/src/Spectre.Algorithms.Tests/Results/DivikResultTests.cs
+++ b/src/Spectre.Algorithms.Tests/Results/DivikResultTests.cs
@@ -69,15 +69,7 @@ namespace Spectre.Algorithms.Tests.Results
             string jsonData = File.ReadAllText(path);
             DivikResult deserialisedResult = JsonConvert.DeserializeObject<DivikResult>(jsonData);
 
-            Assert.AreEqual(_result.AmplitudeFilter, deserialisedResult.AmplitudeFilter, "The AmplitudeFilter differs from original");
-            Assert.AreEqual(_result.AmplitudeThreshold, deserialisedResult.AmplitudeThreshold, "The AmplitudeThreshold differs from original");
-            Assert.AreEqual(_result.Centroids, deserialisedResult.Centroids, "The Centroids differ from original");
-            Assert.AreEqual(_result.Merged, deserialisedResult.Merged, "The Merged differs from original");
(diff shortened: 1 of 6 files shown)


src/Spectre.Algorithms/Results/DivikResult.cs
```diff
@@ -19,63 +19,68 @@
...
+    /// </summary>
+    public sealed class DivikResult
     {
         #region Properties
+
         /// <summary>
         /// Clustering quality index
         /// </summary>
-        public double QualityIndex { get; }
+        public double QualityIndex { get; set; }
```
gmrukwa: Does this library require *public* setters?

src/Spectre.Algorithms/Results/DivikResult.cs
```diff
@@ -93,9 +98,19 @@ internal DivikResult(object matlabResult)
             Merged = ParseMerged(array);
             Subregions = ParseSubregions(array);
         }
+
+        /// <summary>
+        /// Empty constructor for serialization purposes.
+        /// </summary>
+        public DivikResult()
```
gmrukwa: Is this constructor required to be *public*?

src/Spectre.Algorithms/Results/DivikResult.cs
```diff
@@ -216,18 +231,24 @@ private DivikResult[] ParseSubregions(MWStructArray array)
...
         /// <summary>
         /// Saves result under the specified path in JSON format.
         /// </summary>
         /// <param name="path">The destination path.</param>
-        /// <exception cref="System.NotImplementedException">Always, as result saving has not been implemented yet.</exception>
         public void Save(string path)
         {
-            throw new NotImplementedException("Result saving has not been implemented yet.");
+            string data = JsonConvert.SerializeObject(this);
+            using (StreamWriter sw = File.CreateText(path))
```
gmrukwa: Do we need `StreamWriter`? Maybe just `File.WriteAllText(path, data)`?

src/Spectre.Algorithms/Results/DivikResult.cs
Sand3r-: Yes.

Sand3r-: Unlike I thought (I blame it on java) it can actually be even _private_. I'll change that.

src/Spectre.Algorithms/Results/DivikResult.cs
Sand3r-: File.WriteAllText would certainly fit better here. This is actually something that I've been looking for, but decided to pick StreamWriter because of lack of time and as that was the first thing that I came across. 
There indeed are exceptions that are thrown by both of these, and I was truly wondering whether I should list all of them here, since there are a plenty of them and their descriptions would just reproduce those found in CreateText method.

Sand3r-: Nevertheless, The WriteAllText is just a wrapper around the StreamWriter so it just introduces additional overhead. Do we need such change?

src/Spectre.Algorithms/Results/DivikResult.cs
FreexD: +1

src/Spectre.Algorithms/Results/DivikResult.cs
FreexD: We stick to File. in testing, do imho let's nr consistent ;)

FreexD: So* and be* . Sry

src/Spectre.Algorithms/Results/DivikResult.cs
Sand3r-: Yes.

src/Spectre.Algorithms/Results/DivikResult.cs
Sand3r-: Done

src/Spectre.Algorithms.Tests/Results/DivikResultTests.cs
```diff
@@ -52,7 +54,25 @@ public void SetUpFixture()
...
+            Assert.AreEqual(_result.AmplitudeFilter, deserialisedResult.AmplitudeFilter, "The AmplitudeFilter differs from original");
+            Assert.AreEqual(_result.AmplitudeThreshold, deserialisedResult.AmplitudeThreshold, "The AmplitudeThreshold differs from original");
+            Assert.AreEqual(_result.Centroids, deserialisedResult.Centroids, "The Centroids differ from original");
+            Assert.AreEqual(_result.Merged, deserialisedResult.Merged, "The Merged differs from original");
+            Assert.AreEqual(_result.Partition, deserialisedResult.Partition, "The Partition differs from original");
+            Assert.AreEqual(_result.QualityIndex, deserialisedResult.QualityIndex, "The QualityIndex differs from original");
+            Assert.AreEqual(_result.VarianceFilter, deserialisedResult.VarianceFilter, "The VarianceFilter differs from original");
+            Assert.AreEqual(_result.VarianceThreshold, deserialisedResult.VarianceThreshold, "The VarianceThreshold differs from original");
+            Assert.AreEqual(_result.Subregions, deserialisedResult.Subregions, "The Subregions differ from original");
+            File.Delete(path);
```
gmrukwa: Shouldn't this part be removed in `finally` clause? Or do we want to preserve any information (i.e. in the form of created JSON file)?

Issue #61, opened 2017-03-22
Implement StartDivikButtonExecute command


gmrukwa:
The task is to replace StartDivikButtonExecute dummy implementation with the proper one (after all algorithms are configured).

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Implement StartDivikButtonExecute command](https://trello.com/c/I7Bg8PMF/79-implement-startdivikbuttonexecute-command)

gmrukwa:
Closed as 7c5dcb8

Issue #58, opened 2017-03-22
Add license headers to WPF client


gmrukwa:

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Add license headers to WPF client](https://trello.com/c/RPcNmmhd/75-add-license-headers-to-wpf-client)

gmrukwa:
Closed as 7c5dcb8

Issue #57, opened 2017-03-22
Handle validation of input in WPF client


gmrukwa:
The task is to properly handle attribute-based validation of inputs.

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Handle validation](https://trello.com/c/SSYPXPwi/78-handle-validation)

gmrukwa:
Closed as 7c5dcb8

Issue #53, opened 2017-03-19
Provide configuration of DiviK in WPF client


gmrukwa:
The task is to mirror all configurations available in DiviK in WPF client.

gmrukwa:
![](https://github.trello.services/images/mini-trello-icon.png) [Provide configuration of DiviK in WPF client](https://trello.com/c/AdkevJ9p/73-provide-configuration-of-divik-in-wpf-client)

gmrukwa:
Closed as 7c5dcb8

PR #72, opened 2017-03-30
WPF client

gmrukwa:
This branch contains whole WPF client for DiviK algorithm.

Resolves: #61, #58, #57, #53, #52, #47.


gmrukwa:
@Sand3r- honestly, MainPageVm also should be tested - that's one of the purposes of using MVVM pattern. But I would leave it as it is, since it is one-time app. We will proceed with web client further on rather than this solution.

Sand3r-:
Alrighty, feel free to merge then.

gmrukwa:
[Prepare standalone DiviK software](https://trello.com/c/CPUL9yxs/63-prepare-standalone-divik-software)

gmrukwa:
[Logging in matlab/cpp](https://trello.com/c/9VL91hXk/88-logging-in-matlab-cpp)

PR #81, opened 2017-04-23
Move Segmentation to CLI

### LINKED ISSUE #80: Move Segmentation class to C++/CLI
The task is to move the class providing segmentation to C++/CLI (or move some components under the hood to C++/CLI), to be able to call native functions, while having consistent interface outside and possibility to call MATLAB still.

gmrukwa:
Moves Segmentation class to C++/CLI. For now it is just migration to another project, implementation itself is absolutely the same, just in managed C++.

**Note:** *Some changes in access modifiers had to be made, due to extraction of a single class to another project. They must be restored after all algorithms are migrated into C++.*

Resolves #80.


gmrukwa:
[Move Segmentation class to C++/CLI](https://trello.com/c/DFP8wAMG/97-move-segmentation-class-to-c-cli)

PR #84, opened 2017-05-11
GmmResult wrapper and GmmFiltering

gmrukwa:
Purpose of this branch was:
* to extract Gaussian Mixture Model-based filtration outside, to further use it in DiviK (coupled with native k-means algorithm implementation)
* to provide an implementation of GmmResult to make it possible to implement natively threshold finding

As some strange things were spotted at MATLAB site, GMM-modelling has been split into several independent parts and interfaced.

### DIFF
--- a/libs/Debug/MatlabAlgorithmsNative.xml
+++ b/libs/Debug/MatlabAlgorithmsNative.xml
@@ -9,6 +9,9 @@
             The Preprocessing class provides a CLS compliant, Object (native) interface to the
             MATLAB functions contained in the files:
             <newpara></newpara>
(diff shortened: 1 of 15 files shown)


libs/Debug/MatlabAlgorithmsNative.xml
```diff
@@ -9,6 +9,9 @@
             The Preprocessing class provides a CLS compliant, Object (native) interface to the
             MATLAB functions contained in the files:
             <newpara></newpara>
+            C:\Users\Grzegorz\Source\Repos\MSI-dev\spectre\src\MATLAB\DiviK\libs\miscellaneous\f
```
Sebaxp95: Should we use our folders here?

PR #88, opened 2017-06-30
Glip integration

gmrukwa:
This branch introduces Glip integration. On each build, a notification will be sent to Continuous Integration channel, with some basic info about the build.

Versioning schema has been updated - we will have to stick to it and increase version numbers manually on each feature and hotfix.

Bug with C# `ConsoleCaptureService` should be resolved through additional waiting before disposing the service.

*Note*: This branch should be **squashed and merged** at the end.

### DIFF
--- a/src/Spectre.Service/TooSmallUpdateIntervalException.cs
+++ b/src/Spectre.Service/TooSmallUpdateIntervalException.cs
@@ -1,0 +1,28 @@
+﻿using System;
+
+namespace Spectre.Service
+{
+    /// <summary>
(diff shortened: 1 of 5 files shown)


src/Spectre.Service/TooSmallUpdateIntervalException.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
```
Daroslaw: Missing license header

gmrukwa: Good point.

PR #89, opened 2017-06-30
Feature/native partition

Daroslaw:
Implemented Partition utility in C++ (Spectre.libClustering), also added test project based on GoogleTest (Spectre.libClustering.Tests). Proper scripts are also included to make AppVeyor handle the tests.

### DIFF
--- a/src/Common/Main.cpp
+++ b/src/Common/Main.cpp
@@ -1,6 +1,6 @@
 #define GTEST_LANG_CXX11 1
 
-#include <gtest\gtest.h>
+#include <gtest/gtest.h>
 
 int main(int argc, char **argv) 
 {
--- a/src/Common/Main.cpp
+++ b/src/Common/Main.cpp
@@ -6,6 +6,5 @@ int main(int argc, char **argv)
 {
 	::testing::InitGoogleTest(&argc, argv);
 	int testsResult = RUN_ALL_TESTS();
-	system("PAUSE");
 	return testsResult;
 }
--- a/src/Spectre.libClustering.Tests/PartitionTest.cpp
+++ b/src/Spectre.libClustering.Tests/PartitionTest.cpp
@@ -41,7 +41,7 @@ namespace Spectre::libClustering::Tests
 
 	TEST_F(PartitionTest, simplify_sorts_labels)
 	{
-		Partition test(testData);
+		Partition test(std::vector<unsigned int>({ 5, 6, 8, 4, 1, 2, 7, 3 }));
 		auto result = test.Get();
 		ASSERT_TRUE(std::is_sorted(result.begin(), result.end()))
 			<< "Simplification does not sort the labels.";
--- a/src/Spectre.libClustering.Tests/PartitionTest.cpp
+++ b/src/Spectre.libClustering.Tests/PartitionTest.cpp
@@ -5,6 +5,7 @@
 #include <gsl.h>
 #include <vector>
 #include <algorithm>
+#include <exception>
 #include "Partition.h"
 
 namespace Spectre::libClustering::Tests
@@ -22,15 +23,15 @@ namespace Spectre::libClustering::Tests
 
 	};
 
-	TEST_F(PartitionTest, NotFailsOnNull)
+	TEST_F(PartitionTest, fails_on_null)
 	{
-		ASSERT_NO_FATAL_FAILURE({
+		ASSERT_THROW({
 			Partition test(nullptr);
-		}) 
-			<< "Creation from null data caused fatal failure.";
+		}, std::invalid_argument) 
+			<< "Creation from null data did not throw exception.";
 	}
 
-	TEST_F(PartitionTest, SimplifyToNotEmpty)
+	TEST_F(PartitionTest, simplify_to_not_empty)
 	{
 		Partition test(testData);
 		auto result = test.Get();
@@ -38,7 +39,7 @@ namespace Spectre::libClustering::Tests
 			<< "Sequence has been simplified to empty partition.";
 	}
 
-	TEST_F(PartitionTest, SimplifySortsLabels)
+	TEST_F(PartitionTest, simplify_sorts_labels)
 	{
 		Partition test(testData);
 		auto result = test.Get();
@@ -46,7 +47,7 @@ namespace Spectre::libClustering::Tests
 			<< "Simplification does not sort the labels.";
 	}
 
-	TEST_F(PartitionTest, ComparisonDifferentLengths)
+	TEST_F(PartitionTest, comparison_different_lengths)
 	{
 		Partition test1(testData);
 		Partition test2(std::vector<unsigned int>(testData.begin(), testData.end() - 1));
@@ -54,14 +55,14 @@ namespace Spectre::libClustering::Tests
 			<< "Marked equal on partition length mismatch.";
 	}
 
-	TEST_F(PartitionTest, ComparisonEqualPartitionsNoTolerance)
+	TEST_F(PartitionTest, comparison_equal_partitions_no_tolerance)
 	{
 		Partition test(testData);
 		ASSERT_EQ(test, test)
 			<< "Marked not equal on comparison of the same instance.";
 	}
 
-	TEST_F(PartitionTest, ComparisonNotEqualPartitionsNoTolerance)
+	TEST_F(PartitionTest, comparison_not_equal_partitions_no_tolerance)
 	{
 		Partition test1(testData);
 		std::vector<unsigned int> testData2 = { 1, 2, 3, 3, 4, 5, 6, 7 };
@@ -70,7 +71,7 @@ namespace Spectre::libClustering::Tests
 			<< "Marked equal on comparison of different data.";
 	}
 
-	TEST_F(PartitionTest, ComparisonEqualPartitionsMixedLabelsNoTolerance)
+	TEST_F(PartitionTest, comparison_equal_partitions_mixed_labels_no_tolerance)
 	{
 		Partition test1(testData);
 		std::vector<unsigned int> testData2 = { 8, 7, 6, 5, 4, 3, 2, 1 };
@@ -79,7 +80,7 @@ namespace Spectre::libClustering::Tests
 			<< "Comparison is label-sensitive.";
 	}
 
-	TEST_F(PartitionTest, ComparisonEqualPartitionsWithTolerance)
+	TEST_F(PartitionTest, comparison_equal_partitions_with_tolerance)
 	{
 		Partition test1(testData);
 		std::vector<unsigned int> testData2 = testData;
@@ -89,7 +90,7 @@ namespace Spectre::libClustering::Tests
 			<< "Comparison is tolerance insensitive.";
 	}
 
-	TEST_F(PartitionTest, ComparisonNotEqualPartitionsWithTolerance)
+	TEST_F(PartitionTest, comparison_not_equal_partitions_with_tolerance)
 	{
 		Partition test1(testData);
 		std::vector<unsigned int> testData2 = { 2, 3, 3, 5, 5, 7, 7, 9 };
@@ -98,7 +99,7 @@ namespace Spectre::libClustering::Tests
 			<< "Comparison did not capture inequality.";
 	}
 
-	TEST_F(PartitionTest, ComparisonEqualPartitionsMixedLabelsWithTolerance)
+	TEST_F(PartitionTest, comparison_equal_partitions_mixed_labels_with_tolerance)
 	{
 		Partition test1(testData);
 		std::vector<unsigned int> testData2 = { 8, 7, 6, 5, 4, 3, 2, 2 };
--- a/src/Spectre.libClustering.Tests/PartitionTest.cpp
+++ b/src/Spectre.libClustering.Tests/PartitionTest.cpp
@@ -1,12 +1,11 @@
 #define GTEST_LANG_CXX11 1
 
-#include <gtest\gtest.h>
-#include <gtest\gtest-spi.h>
+#include <gtest/gtest.h>
 #include <gsl.h>
 #include <vector>
 #include <algorithm>
-#include <exception>
 #include "Partition.h"
+#include "ArgumentNullException.h"
 
 namespace Spectre::libClustering::Tests
 {
@@ -16,7 +15,7 @@ namespace Spectre::libClustering::Tests
 
 		std::vector<unsigned int> testData;
 
-		virtual void SetUp()
+		virtual void SetUp() override
 		{
 			testData = { 1, 2, 3, 4, 5, 6, 7, 8 };
 		}
@@ -27,7 +26,7 @@ namespace Spectre::libClustering::Tests
 	{
 		ASSERT_THROW({
 			Partition test(nullptr);
-		}, std::invalid_argument) 
+		}, ArgumentNullException) 
 			<< "Creation from null data did not throw exception.";
 	}
 
@@ -41,7 +40,8 @@ namespace Spectre::libClustering::Tests
 
 	TEST_F(PartitionTest, simplify_sorts_labels)
 	{
-		Partition test(std::vector<unsigned int>({ 5, 6, 8, 4, 1, 2, 7, 3 }));
+		auto unsorted = std::vector<unsigned int>({ 5, 6, 8, 4, 1, 2, 7, 3 });
+		Partition test(unsorted);
 		auto result = test.Get();
 		ASSERT_TRUE(std::is_sorted(result.begin(), result.end()))
 			<< "Simplification does not sort the labels.";
@@ -50,7 +50,8 @@ namespace Spectre::libClustering::Tests
 	TEST_F(PartitionTest, comparison_different_lengths)
 	{
 		Partition test1(testData);
-		Partition test2(std::vector<unsigned int>(testData.begin(), testData.end() - 1));
+		auto shorter = std::vector<unsigned int>(testData.begin(), testData.end() - 1);
+		Partition test2(shorter);
 		ASSERT_NE(test1, test2)
 			<< "Marked equal on partition length mismatch.";
 	}
--- a/src/Spectre.libClustering.Tests/PartitionTest.cpp
+++ b/src/Spectre.libClustering.Tests/PartitionTest.cpp
@@ -1,0 +1,110 @@
+#define GTEST_LANG_CXX11 1
+
+#include <gtest\gtest.h>
+#include <gtest\gtest-spi.h>
+#include <gsl.h>
+#include <vector>
+#include <algorithm>
+#include "Partition.h"
+
+namespace Spectre::libClustering::Tests
+{
+	class PartitionTest : public ::testing::Test
+	{
+	protected:
+
+		std::vector<unsigned int> testData;
+
+		virtual void SetUp()
+		{
+			testData = { 1, 2, 3, 4, 5, 6, 7, 8 };
+		}
+
+	};
+
+	TEST_F(PartitionTest, NotFailsOnNull)
+	{
+		ASSERT_NO_FATAL_FAILURE({
+			Partition test(nullptr);
+		}) 
+			<< "Creation from null data caused fatal failure.";
+	}
+
+	TEST_F(PartitionTest, SimplifyToNotEmpty)
+	{
+		Partition test(testData);
+		auto result = test.Get();
+		ASSERT_NE(result.size(), 0)
+			<< "Sequence has been simplified to empty partition.";
+	}
+
+	TEST_F(PartitionTest, SimplifySortsLabels)
+	{
+		Partition test(testData);
+		auto result = test.Get();
+		ASSERT_TRUE(std::is_sorted(result.begin(), result.end()))
+			<< "Simplification does not sort the labels.";
+	}
+
+	TEST_F(PartitionTest, ComparisonDifferentLengths)
+	{
+		Partition test1(testData);
+		Partition test2(std::vector<unsigned int>(testData.begin(), testData.end() - 1));
+		ASSERT_NE(test1, test2)
+			<< "Marked equal on partition length mismatch.";
+	}
+
+	TEST_F(PartitionTest, ComparisonEqualPartitionsNoTolerance)
+	{
+		Partition test(testData);
+		ASSERT_EQ(test, test)
+			<< "Marked not equal on comparison of the same instance.";
+	}
+
+	TEST_F(PartitionTest, ComparisonNotEqualPartitionsNoTolerance)
+	{
+		Partition test1(testData);
+		std::vector<unsigned int> testData2 = { 1, 2, 3, 3, 4, 5, 6, 7 };
+		Partition test2(testData2);
+		ASSERT_NE(test1, test2)
+			<< "Marked equal on comparison of different data.";
+	}
+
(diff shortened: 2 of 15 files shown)


src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+		{
+			testData = { 1, 2, 3, 4, 5, 6, 7, 8 };
+		}
+
+	};
+
+	TEST_F(PartitionTest, NotFailsOnNull)
+	{
+		ASSERT_NO_FATAL_FAILURE({
+			Partition test(nullptr);
```
gmrukwa: I think this one should be disallowed.

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+		ASSERT_NE(result.size(), 0)
+			<< "Sequence has been simplified to empty partition.";
+	}
+
+	TEST_F(PartitionTest, SimplifySortsLabels)
+	{
+		Partition test(testData);
+		auto result = test.Get();
+		ASSERT_TRUE(std::is_sorted(result.begin(), result.end()))
+			<< "Simplification does not sort the labels.";
```
gmrukwa: Order between sorted and unsorted does not differ. This may hide error.

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+		std::vector<unsigned int> testData;
+
+		virtual void SetUp()
+		{
+			testData = { 1, 2, 3, 4, 5, 6, 7, 8 };
+		}
+
+	};
+
+	TEST_F(PartitionTest, NotFailsOnNull)
```
gmrukwa: I would go for more readable names of tests like `not_fails_on_null`.

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+		Partition test2(testData2);
+		ASSERT_EQ(test1, test2)
+			<< "Comparison is label-sensitive.";
+	}
+
+	TEST_F(PartitionTest, ComparisonEqualPartitionsWithTolerance)
+	{
+		Partition test1(testData);
+		std::vector<unsigned int> testData2 = testData;
+		testData2.back() -= 1;
```
gmrukwa: This will have to be extended at some point, to grasp permutations and their best compatibility.

src/Spectre.libClustering.Tests/Spectre.libClustering.Tests.vcxproj
```diff
@@ -0,0 +1,171 @@
...
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <LinkIncremental>true</LinkIncremental>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
gmrukwa: Win32?

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,73 @@
...
+	Partition::Partition(gsl::span<unsigned int> partition)
+	{
+		std::map<unsigned int, int> labelsDictionary;
+
+		int currentIntLabel = 1;
+		for (auto label : partition)
+			if (labelsDictionary.find(label) == labelsDictionary.end())
+				labelsDictionary.emplace(label, currentIntLabel++);
+
+		m_Partition.reserve(partition.size());
```
gmrukwa: You can do it in an initialization list.

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,73 @@
...
+		auto rhsPartition = rhs.Get();
+
+		if (lhsPartition.size() != rhsPartition.size())
+			return false;
+
+		unsigned int matchesCount = 0;
+		size_t length = lhsPartition.size();
+
+		for (int i = 0; i < length; i++)
+			if (lhsPartition.at(i) == rhsPartition.at(i))
```
gmrukwa: Cannot normal `operator[]` be used? It should prevent us from modification without knowing, but allow to read.

src/Spectre.libClustering/Partition.h
```diff
@@ -0,0 +1,78 @@
...
+		/// <param name="rhs">The second partition.</param>
+		/// <param name="tolerance">The tolerance rate of mismatch.</param>
+		static bool Compare(const Partition &lhs, const Partition &rhs, double tolerance = 0);
+
+		/// <summary>
+		/// Overloaded equality operator. Compares with tolerance = 0.
+		/// </summary>
+		/// <param name="lhs">The first partition.</param>
+		/// <param name="rhs">The second partition.</param>
+		friend bool operator== (const Partition &lhs, const Partition &rhs);
```
gmrukwa: Does it have to be friend? Shouldn't it be const also?

src/Spectre.libClustering/Partition.h
```diff
@@ -0,0 +1,78 @@
...
+
+		/// <summary>
+		/// Overloaded inequality operator. Compares with tolerance = 0.
+		/// </summary>
+		/// <param name="lhs">The first partition.</param>
+		/// <param name="rhs">The second partition.</param>
+		friend bool operator!= (const Partition &lhs, const Partition &rhs);
```
gmrukwa: Does it have to be friend? Shouldn't it be const also?

src/Spectre.libClustering/Partition.h
```diff
@@ -0,0 +1,78 @@
...
+
+	private:
+		std::vector<unsigned int> m_Partition;
```
gmrukwa: Do we ever allow to change it? `const` (or other `readonly`)? Or such constructs work only in C#?

src/Spectre.libClustering/Spectre.libClustering.vcxproj
```diff
@@ -0,0 +1,153 @@
...
+      <WarningLevel>Level3</WarningLevel>
+      <Optimization>Disabled</Optimization>
+      <PreprocessorDefinitions>_DEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <AdditionalOptions>/std:c++latest %(AdditionalOptions)</AdditionalOptions>
+    </ClCompile>
+    <Link>
+      <SubSystem>Windows</SubSystem>
+    </Link>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
gmrukwa: Win32

src/Spectre.libClustering.Tests/PartitionTest.cpp
Sand3r-: They look alright to me as they are now

src/Spectre.libClustering/Partition.cpp
Sand3r-: .reserve() in an initialization list?

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,73 @@
...
+		std::map<unsigned int, int> labelsDictionary;
+
+		int currentIntLabel = 1;
+		for (auto label : partition)
+			if (labelsDictionary.find(label) == labelsDictionary.end())
+				labelsDictionary.emplace(label, currentIntLabel++);
+
+		m_Partition.reserve(partition.size());
+		for (int i = 0; i < partition.size(); i++)
+			m_Partition.push_back(labelsDictionary[partition[i]]);
```
Sand3r-: If the change proposed by Grzesiek is implemented, push_back needs to be replaces with traditional array access.

src/Spectre.libClustering/Partition.cpp
Sand3r-: Since Get() returns const reference it is true, nevertheless personally I don't see anything wrong with using .at(i).

src/Spectre.libClustering/Partition.h
Sand3r-: That's not possible since we fill it with values in constructor's body, not in initialization list.

src/Spectre.libClustering/Partition.h
Daroslaw: Friend is needed so the operator is accessible from the global scope without the need for i.e. GoogleTest to know about something called "Partition". This makes it automatically a non-member function (free), so qualifiers like "const" are not allowed.

src/Spectre.libClustering/Partition.h
Daroslaw: @up

src/Spectre.libClustering/Partition.h
Sand3r-: Afaik, it needs to be friend, since it uses private Partition members.

src/Spectre.libClustering/Partition.cpp
Daroslaw: AFAIK the `operator[]` only differs from `.at()` in the fact that `.at()` does some runtime range checks. I will change it to `operator[]` for the sick speeds but the data is unmodifiable in both cases.

src/Spectre.libClustering/Partition.cpp
Daroslaw: What are the benefits of this?

src/Spectre.libClustering/Partition.h
Daroslaw: It is modified extensively only in the constructor but that is the main reason it should be non-constant.

src/Spectre.libClustering/Partition.cpp
Sand3r-: Currently std::vector's constructor is called upon entering Partition's constructor body, and then we perform another call to reserve memory, while this can be solved by a single constructor call in the initialization list.

src/Spectre.libClustering/Partition.cpp
Sand3r-: at() also requires instruction jump and return on each call, while using index operator only requires 0-cost address translation

src/Spectre.libClustering.Tests/PartitionTest.cpp
Daroslaw: I changed to the variant proposed by Grzegorz. Comparing names like:
`comparison_equal_partitions_mixed_labels_no_tolerance` with
`comparison_equal_partitions_mixed_labels_with_tolerance`
seems to be easier than:
`ComparisonEqualPartitionsMixedLabelsNoTolerance` with
`ComparisonEqualPartitionsMixedLabelsWithTolerance`

src/Spectre.libClustering.Tests/PartitionTest.cpp
Daroslaw: Done

src/Spectre.libClustering.Tests/PartitionTest.cpp
Daroslaw: Could you provide some additional information?

src/Spectre.libClustering.Tests/PartitionTest.cpp
Daroslaw: Could you provide some additional information?

src/Spectre.libClustering.Tests/Spectre.libClustering.Tests.vcxproj
Daroslaw: Deleted

src/Spectre.libClustering/Spectre.libClustering.vcxproj
Daroslaw: Deleted

src/Spectre.libClustering/Partition.cpp
Daroslaw: Done

src/Spectre.libClustering.Tests/PartitionTest.cpp
Sand3r-: The problem is that the testData is already sorted, so it's hard to test if constructor has sorted the labels.

src/Spectre.libClustering/Partition.h
Sand3r-: According to @gmrukwa if we remove the first argument (const Partition &lhs) the friend keyword can be omitted because it would no longer serve as an operator overload but as a member function overload.

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "Partition.h"
+#include <exception>
+
+namespace Spectre::libClustering
+{
+	Partition::Partition(gsl::span<unsigned int> partition) : m_Partition(partition.size())
```
Sand3r-: From @gmrukwa : please add "explicit" keyword to the constructor to avoid implicit conversions.

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+			return false;
+
+		unsigned int matchesCount = 0;
+		size_t length = lhsPartition.size();
+
+		for (int i = 0; i < length; i++)
+			if (lhsPartition[i] == rhsPartition[i])
+				++matchesCount;
+
+		double compatibilityRate = matchesCount / (double)length;
```
gmrukwa: Please use `static_cast`.

src/Spectre.libClustering/Partition.h
gmrukwa: It works well with GTest, try this:
```
bool operator==(const Partition& other) const
        {
            return Compare(*this, other);
        }

        bool operator!=(const Partition& other) const
        {
            return !((*this) == other);
        }
```

gmrukwa: E.g. it introduces `const` which provides more control.

src/Spectre.libClustering/Partition.h
gmrukwa: up

src/Spectre.libClustering/Partition.h
gmrukwa: K - I thought about C#-like `readonly`

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+namespace Spectre::libClustering
+{
+	Partition::Partition(gsl::span<unsigned int> partition) : m_Partition(partition.size())
+	{
+		if (partition.data() == nullptr)
+			throw std::invalid_argument("The input partition is null");
+
+		std::map<unsigned int, int> labelsDictionary;
+
+		int currentIntLabel = 1;
```
gmrukwa: auto

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+		for (auto label : partition)
+			if (labelsDictionary.find(label) == labelsDictionary.end())
+				labelsDictionary.emplace(label, currentIntLabel++);
+
+		for (int i = 0; i < partition.size(); i++)
```
gmrukwa: auto

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+	bool Partition::Compare(const Partition & lhs, const Partition & rhs, double tolerance)
+	{
+		auto& lhsPartition = lhs.Get();
+		auto& rhsPartition = rhs.Get();
+
+		if (lhsPartition.size() != rhsPartition.size())
+			return false;
+
+		unsigned int matchesCount = 0;
+		size_t length = lhsPartition.size();
```
gmrukwa: auto

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+		for (int i = 0; i < length; i++)
```
gmrukwa: auto

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+			if (lhsPartition[i] == rhsPartition[i])
+				++matchesCount;
+
+		double compatibilityRate = matchesCount / static_cast<double>(length);
```
gmrukwa: auto

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+		double requiredCompatibilityRate = 1 - tolerance;
```
gmrukwa: auto

src/Spectre.libClustering/Partition.h
```diff
@@ -0,0 +1,76 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#pragma once
+
+#include <vector>
+#include <map>
```
gmrukwa: This one should be in `.cpp` file.

src/Spectre.libClustering/Partition.h
```diff
@@ -0,0 +1,76 @@
...
+#include <algorithm>
```
gmrukwa: Is this necessary?

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "Partition.h"
+#include <exception>
```
gmrukwa: This is unnecessary.

src/Spectre.libClustering/Partition.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Spectre::libClustering
+{
+	Partition::Partition(gsl::span<unsigned int> partition) : m_Partition(partition.size())
+	{
+		if (partition.data() == nullptr)
+			throw std::invalid_argument("The input partition is null");
```
gmrukwa: Don't we introduce custom exception?

gmrukwa: Message should provide information about context, no user-oriented message.

I.e.:
`throw ArgumentNullException("partition")` <- uses name of the function input which is null
instead of:
`throw std::invalid_argument("The input partition is null")` <- introduces problems during logging, which will have to be introduced at some point

Of course, such `ArgumentNullException` may be a subclass of `std::invalid_argument`. At some point we will introduce our own exception base, so all exceptions per project can be in single file for now (to change exception base quickly at that point).

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+	{
+		Partition test(testData);
+		auto result = test.Get();
+		ASSERT_NE(result.size(), 0)
+			<< "Sequence has been simplified to empty partition.";
+	}
+
+	TEST_F(PartitionTest, simplify_sorts_labels)
+	{
+		Partition test(std::vector<unsigned int>({ 5, 6, 8, 4, 1, 2, 7, 3 }));
```
gmrukwa: Static checks cry for missing move constructor.

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+		auto result = test.Get();
+		ASSERT_TRUE(std::is_sorted(result.begin(), result.end()))
+			<< "Simplification does not sort the labels.";
+	}
+
+	TEST_F(PartitionTest, comparison_different_lengths)
+	{
+		Partition test1(testData);
+		Partition test2(std::vector<unsigned int>(testData.begin(), testData.end() - 1));
```
gmrukwa: Static checks cry for missing move constructor.

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Spectre::libClustering::Tests
+{
+	class PartitionTest : public ::testing::Test
+	{
+	protected:
+
+		std::vector<unsigned int> testData;
+
+		virtual void SetUp()
```
gmrukwa: Missing `override` specifier.

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,111 @@
+#define GTEST_LANG_CXX11 1
+
+#include <gtest\gtest.h>
+#include <gtest\gtest-spi.h>
```
gmrukwa: Is this necessary?

src/Spectre.libClustering.Tests/PartitionTest.cpp
```diff
@@ -0,0 +1,111 @@
+#define GTEST_LANG_CXX11 1
+
+#include <gtest\gtest.h>
```
gmrukwa: Use forward slash for compatibility.

src/Common/Main.cpp
```diff
@@ -0,0 +1,10 @@
+#define GTEST_LANG_CXX11 1
+
+#include <gtest\gtest.h>
+
+int main(int argc, char **argv) 
+{
+	::testing::InitGoogleTest(&argc, argv);
+	int testsResult = RUN_ALL_TESTS();
```
gmrukwa: auto

src/Common/Main.cpp
```diff
@@ -0,0 +1,10 @@
+#define GTEST_LANG_CXX11 1
+
+#include <gtest\gtest.h>
```
gmrukwa: Use forward slashes for compatibility.

src/Spectre.libClustering/Partition.cpp
gmrukwa: This has been already done in `.h` file, but comment did not go outdated.

src/Spectre.libClustering/Partition.cpp
Sand3r-: No.

gmrukwa: xD

PR #90, opened 2017-07-02
Increase warning level

gmrukwa:
Increases warning level and starts to treat warnings as errors.
Fixes all warning-generating places.

This should improve quality of code. Warning level and
treat warnings as errors should be enabled in any new project.


gmrukwa:
[Increase warning level](https://trello.com/c/52DIZIGM/151-increase-warning-level)

PR #91, opened 2017-07-03
Feature/gmm model native

Sand3r-:
Feature introduces the Gaussian Mixture Model structure that will serve as return type generated by GMM-family algorithms.


src/Spectre.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26430.12
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
```
mg6: Not necessarily needed. Watch out for this! :)

Seems fine otherwise.

Sand3r-:
[Native GmmModel](https://trello.com/c/TwTrkyml/134-native-gmmmodel)

src/Spectre.libGmmModel/GaussianMixtureModel.h
```diff
@@ -0,0 +1,82 @@
...
+	/// <summary>
+	/// Represents a composite of gaussian distribution components 
+	/// that build up a Gaussian Mixture Model.
+	/// </summary>
+	struct GaussianMixtureModel
+	{
+		/// <summary>
+		/// Collection of Gaussian components
+		/// </summary>
+		gsl::span<GaussianComponent> components;
```
gmrukwa: This is all fine, glowing and shiny, but are we providing any possibility `GaussianMixtureModel` will *own* `GaussianComponent` instances? Maybe there should be a private e.g. `std::vector` field which will hold the data for the object lifespan, while `components` will just point to this?
**In general**: we want to *accept* `gsl::span`, to provide flexibility of *input*, while it is usually useful to have quite specific *output*. That's the concept I see in using `gsl::span` - flexibility of input, and well-specified output.

src/Spectre.libGmmModel/GaussianMixtureModel.h
```diff
@@ -0,0 +1,82 @@
...
+
+		/// <summary>
+		/// Collection of initially supplied mz values
+		/// </summary>
+		gsl::span<double> originalMzArray;
```
gmrukwa: As above.

src/Spectre.libGmmModel/GaussianMixtureModel.h
```diff
@@ -0,0 +1,82 @@
...
+
+		/// <summary>
+		/// Collection of average supplied spectra.
+		/// </summary>
+		gsl::span<double> originalMeanSpectrum;
```
gmrukwa: As above.

src/Spectre.libGmmModel/GaussianMixtureModel.h
Sand3r-: I thought this, and originalMeanSpectrum are given from elsewhere upfront.

gmrukwa: We have to think about this model as of serializable and standalone unit, which may be saved and restored at any time, without the original dataset, etc.

src/Spectre.libGmmModel/GaussianMixtureModel.h
```diff
@@ -0,0 +1,104 @@
...
+        /// Constructor used for initialization of m/z, intesities and gaussian components
+        /// collections.
+        /// </summary>
+        /// <param name="mzArray">M/z data shared by all spectra.</param>
+        /// <param name="intensities">Mean intensities at each point.</param>
+        /// <param name="numberOfComponents">Number of Gaussian Components to be set.</param>
+        GaussianMixtureModel(const std::vector<double>& mzArray,
+                             const std::vector<double>& intensities,
+                             unsigned int numberOfComponents) :
+            m_OriginalMzArray(mzArray), m_OriginalMeanSpectrum(intensities),
```
gmrukwa: This one won't work either. `m_OriginalMzArray` will be copy of `mzArray`, while `originalMzArray` will be initialized with not initialized `std::vector<double>`. In the initialization list variables are initialized in the order of *declaration in class*, **not** the order we write in *constructor declaration* ([source](https://stackoverflow.com/a/4037283/3067956)).
To achieve what you want, you have to populate those instances in the body of constructor. They aren't `const`, so no problem.

src/Spectre.libGaussianMixtureModelling.Tests/GaussianMixtureModelTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+    TEST_F(GaussianMixtureModelTest, test_data_sizes)
+    {
+        constexpr unsigned int numberOfComponents = 5;
+        GaussianMixtureModel model(testData, testData, numberOfComponents);
+        ASSERT_EQ(model.components.size(), numberOfComponents);
+        ASSERT_EQ(model.originalMeanSpectrum.size(), testData.size());
+        ASSERT_EQ(model.originalMzArray.size(), testData.size());
+    }
+
+    TEST_F(GaussianMixtureModelTest, test_data_correctness)
```
gmrukwa: Please provide more descriptive name (i.e. which data are checked to be correct).

src/Spectre.libGaussianMixtureModelling.Tests/GaussianMixtureModelTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+        {
+            testData = { 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0 };
+        }
+    };
+
+    TEST_F(GaussianMixtureModelTest, test_data_sizes)
+    {
+        constexpr unsigned int numberOfComponents = 5;
+        GaussianMixtureModel model(testData, testData, numberOfComponents);
+        ASSERT_EQ(model.components.size(), numberOfComponents);
```
gmrukwa: `ASSERT_EQ` breaks execution on failure. Was this intended? If not, please use `EXPECT_EQ` instead.

src/Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Collection of average supplied spectra.
+        /// </summary>
+        const std::vector<double> originalMeanSpectrum;
+
+        /// <summary>
+        /// M/z threshold used in components merging.
+        /// </summary>
+        double mzMergingThreshold;
```
gmrukwa: `const`?

src/Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Value indicating whether this instance is merged.
+        /// </summary>
+        bool isMerged;
```
gmrukwa: `const`?

src/Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether this instance is noise components reduced.
+        /// </summary>
+        bool isNoiseReduced;
```
gmrukwa: `const`?

PR #94, opened 2017-07-11
Generic dataset in C++

mg6:
This is an initial implementation of https://trello.com/c/wgOtKAlC/129-generic-dataset-in-c


gmrukwa:
[Generic Dataset in C++](https://trello.com/c/wgOtKAlC/129-generic-dataset-in-c)

src/Spectre.libDataset.Tests/targetver.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+// Including SDKDDKVer.h defines the highest available Windows platform.
+
+// If you wish to build your application for a previous Windows platform, include WinSDKVer.h and
+// set the _WIN32_WINNT macro to the platform you wish to support before including SDKDDKVer.h.
+
+#include <SDKDDKVer.h>
```
gmrukwa: Do we need this? It violates portability.

src/Spectre.libDataset/IDataset.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "IReadOnlyDataset.h"
+
+namespace Spectre::libDataset
+{
+    template <typename DataType, typename SampleMetadata, typename DatasetMetadata>
+    class IDataset : public IReadOnlyDataset<DataType, SampleMetadata, DatasetMetadata>
+    {
+    public:
+        using iterator = typename gsl::span<DataTypeDT>::iterator;
```
gmrukwa: As agreed during talks, we may drop this.

src/Spectre.libDataset/IReadOnlyDataset.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Spectre::libDataset
+{
+    template<typename DataType, typename SampleMetadata, typename DatasetMetadata>
+    class IReadOnlyDataset
+    {
+    public:
+        using DataTypeDT = typename deduce<DataType>::type;
+        using SampleMetadataDT = typename deduce<SampleMetadata>::type;
+        using const_iterator = typename gsl::span<const DataTypeDT>::const_iterator;
```
gmrukwa: As agreed during talks, we may drop this.

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,108 @@
...
+    template<typename DataType, typename SampleMetadata, typename DatasetMetadata>
+    class Dataset : public IDataset<DataType, SampleMetadata, DatasetMetadata>
+    {
+    public:
+        Dataset(gsl::span<const DataTypeDT> data, gsl::span<const SampleMetadataDT> sampleMetadata, DatasetMetadata metadata) :
+            m_data(data), m_sampleMetadata(sampleMetadata), m_metadata(metadata)
+        {
+        }
+
+        virtual const DataTypeDT& operator[](int idx) const override
```
gmrukwa: `virtual` is redundant when `override` is present.

src/Spectre.libDataset.Tests/DatasetTest.cpp
```diff
@@ -0,0 +1,19 @@
+#include "stdafx.h"
+#include "CppUnitTest.h"
```
gmrukwa: Unnecessary, due to the content of `stdafx.h`.

src/Spectre.libDataset/IReadOnlyDataset.h
```diff
@@ -0,0 +1,48 @@
...
+    };
+
+    template<typename T>
+    struct deduce_helper<false, T>
+    {
+        using type = typename const T&;
+    };
+
+    template<typename T>
+    struct deduce : deduce_helper<std::is_fundamental<T>::value || std::is_enum<T>::value || std::is_pointer<T>::value, T>
```
gmrukwa: Seems like this one is destructing [this line](https://github.com/spectre-team/spectre/pull/94/files#diff-94d41ff8ccd535c63a6f4b84f58ebdc3R34). Maybe we should drop it - const ref to e.g. `int` is of the same weight as copy.

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        virtual void setData(gsl::span<const DataTypeDT> data) override
+        {
+            m_data = data;
+        }
+
+        virtual void setSampleMetadata(gsl::span<const SampleMetadataDT> metadata) override
+        {
+            m_metadata = metadata;
```
gmrukwa: `m_sampleMetadata`

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,108 @@
...
+    {
+    public:
+        Dataset(gsl::span<const DataTypeDT> data, gsl::span<const SampleMetadataDT> sampleMetadata, DatasetMetadata metadata) :
+            m_data(data), m_sampleMetadata(sampleMetadata), m_metadata(metadata)
+        {
+        }
+
+        virtual const DataTypeDT& operator[](int idx) const override
+        {
+            return m_data[idx];
```
gmrukwa: Please put function definitions out of the class.

src/Spectre.libDataset/Dataset.h
gmrukwa: Use `assign` member function.

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        virtual gsl::span<const SampleMetadataDT&> getSampleMetadata() const override
+        {
+            return gsl::span<const SampleMetadataDT&>(m_sampleMetadata);
+        }
+
+        virtual void setData(gsl::span<const DataTypeDT> data) override
+        {
+            m_data = data;
```
gmrukwa: Use `assign` member function.

gmrukwa:
@mg6, please provide tests.

src/Spectre.libDataset.Tests/DatasetTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+    data_set dataset(data, sampleMetadata, datasetMetadata);
+}
+
+TEST(DatasetTest, knows_if_empty)
+{
+    samples empty_samples;
+    samples_metadata empty_samples_metadata;
+    dataset_metadata empty_dataset_metadata;
+    data_set dataset(empty_samples, empty_samples_metadata, empty_dataset_metadata);
+    EXPECT_TRUE(dataset.empty());
```
FreexD: Can you prove that for readonly as well?

gmrukwa: `empty` has no overloads for `const` or non-`const` return type. This would make no difference in that case.

FreexD: Lgtm then

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,192 @@
...
+        /// <returns>Dataset metadata</returns>
+        DatasetMetadata& getDatasetMetadata() override
+        {
+            return m_metadata;
+        }
+
+        /// <summary>
+        /// Gets the data in read-only fashion.
+        /// </summary>
+        /// <returns></returns>
```
Sand3r-: Empty docstring

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,192 @@
...
+        gsl::span<const DataType> getData() const override
+        {
+            return gsl::span<const DataType>(m_data);
+        }
+
+        /// <summary>
+        /// Gets the sample metadata in read-only fashion.
+        /// </summary>
+        /// <returns></returns>
```
Sand3r-: See above.

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,192 @@
...
+#include "IDataset.h"
+#include "Exception.h"
+
+namespace Spectre::libDataset
+{
+    /// <summary>
+    /// Data-owning implementation of dataset interfaces.
+    /// </summary>
+    template<typename DataType, typename SampleMetadata, typename DatasetMetadata>
+    class Dataset : public IDataset<DataType, SampleMetadata, DatasetMetadata>
```
Sand3r-: According to [C++ coding convetion](https://github.com/spectre-team/spectre/wiki/Cpp-coding-convention) method names shall start with uppercase letter. This applies to all files created in this commit.

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,192 @@
...
+        /// Checks, whether dataset is empty.
+        /// </summary>
+        /// <returns>True, if dataset is empty.</returns>
+        bool empty() const override
+        {
+            return m_data.empty();
+        }
+
+    private:
+        std::vector<DataType> m_data;
```
Sand3r-: According to [C++ coding convetion](https://github.com/spectre-team/spectre/wiki/Cpp-coding-convention) class members should start with m_ (it's ok) and follow up with an uppercase letter.

src/Spectre.libDataset.Tests/DatasetTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+    { }
+protected:
+    const samples data;
+    const samples_metadata entries_metadata;
+    const dataset_metadata global_metadata;
+    data_set dataset;
+
+    void SetUp() override
+    {
+        dataset = data_set(data, entries_metadata, global_metadata);
```
Sand3r-: Is there any reason behind initializing this both inside initialization list and here?

src/Spectre.libDataset.Tests/DatasetTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+{
+    const auto& local = dataset;
+    EXPECT_THROW(local.getSampleMetadata(100), OutOfRange);
+}
+
+TEST_F(DatasetInScopeTest, access_whole_data_in_readonly)
+{
+    const auto& local = dataset;
+    auto local_data = local.getData();
+    EXPECT_EQ(local_data.size(), 3);
```
Sand3r-: I'd consider using ASSERT_EQ here, since we risk reading out of bound memory in later commands. (Not really harmful, but just a thing to consider).

src/Spectre.libDataset.Tests/DatasetTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+    EXPECT_EQ(local_data[0], 1);
+    EXPECT_EQ(local_data[1], 2);
+    EXPECT_EQ(local_data[2], 3);
+}
+
+TEST_F(DatasetInScopeTest, access_whole_sample_metadata_in_readonly)
+{
+    const auto& local = dataset;
+    auto local_sample_metadata = local.getSampleMetadata();
+    EXPECT_EQ(local_sample_metadata.size(), 3);
```
Sand3r-: Same as above.

src/Spectre.libDataset.Tests/DatasetTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+TEST_F(DatasetInScopeTest, know_size)
+{
+    EXPECT_EQ(dataset.size(), 3);
+}
+
+TEST_F(DatasetInScopeTest, add_new_sample)
+{
+    dataset.addSample(8, 9);
+    EXPECT_EQ(dataset.getData().size(), 4);
+    EXPECT_EQ(dataset.getSampleMetadata().size(), 4);
```
Sand3r-: Here and line below as well.

src/Spectre.libDataset/Dataset.h
```diff
@@ -0,0 +1,192 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Dataset"/> class.
+        /// </summary>
+        /// <param name="data">The data.</param>
+        /// <param name="sampleMetadata">The sample metadata.</param>
+        /// <param name="metadata">The metadata.</param>
+        Dataset(gsl::span<const DataType> data, gsl::span<const SampleMetadata> sampleMetadata, DatasetMetadata metadata) :
+            m_data(data.begin(), data.end()), m_sampleMetadata(sampleMetadata.begin(), sampleMetadata.end()), m_metadata(metadata)
+        {
+            if(data.size() == sampleMetadata.size())
```
gmrukwa: This is for instruction-optimization purposes, like [this](http://igoro.com/archive/fast-and-slow-if-statements-branch-prediction-in-modern-processors/). Compiler assumes, that `if` condition is rather `true` than `false`, so this branch is optimized.

src/Spectre.libDataset/Exception.cpp
```diff
@@ -0,0 +1,45 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "Exception.h"
+
+namespace Spectre::libDataset
+{
+ExceptionBase::ExceptionBase(const std::string message): m_message(message) { }
```
gmrukwa: Added to header file - that's its place.

src/Spectre.libDataset/Exception.h
```diff
@@ -0,0 +1,45 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+#include <string>
+
+namespace Spectre::libDataset
```
gmrukwa: Look above.

src/Spectre.libDataset/Dataset.h
gmrukwa: While I agree on *usual* methods, I think we should leave the interface consistent with *C++-like* whenever mimicking it, so this would apply e.g. to: `size`, `empty`.

src/Spectre.libDataset/Dataset.h
gmrukwa: Fixed.

src/Spectre.libDataset/Dataset.h
gmrukwa: Fixed.

src/Spectre.libDataset/Dataset.h
gmrukwa: Fixed.

src/Spectre.libDataset.Tests/DatasetTest.cpp
gmrukwa: Of course - `SetUp` provides us with unchanged, fresh instance before each of the tests starts, while (due to the lack of default constructor) it had to be initizalized earlier.

src/Spectre.libDataset.Tests/DatasetTest.cpp
gmrukwa: Good point. Accepted.

src/Spectre.libDataset.Tests/DatasetTest.cpp
gmrukwa: Accepted.

src/Spectre.libDataset.Tests/DatasetTest.cpp
gmrukwa: Accepted.

src/Spectre.libDataset/Dataset.h
Sand3r-: The comparison of assembly produced for each case has been presented here:
https://www.diffnow.com/?report=j1bf5
- The proposed solution contains only 2 jumps, while the current 3.
- Moreover, the proposed approach generates less instructions.

Nevertheless:
- Current code generates assembly that during execution of valid dataset goes through a one conditional and one unconditional jump which indeed goes into favor of better branch predictioning.

Therefore, it should indeed stay the way it is in my opinion.

*Checked on O2.*

PR #100, opened 2017-07-27
Move matlab dependency to nuget

gmrukwa:
I've moved our MATLAB dependencies into a NuGet package, so it will no longer be included into our repository.

**Note:** To compile this version, you need an extra setup regarding additional NuGet feed.

[How to get setup?](https://github.com/spectre-team/spectre/wiki/Algorithms-Setup#nuget-feed)


gmrukwa:
[Move dependencies to git LFS](https://trello.com/c/kVG8grJl/82-move-dependencies-to-git-lfs)

gmrukwa:
[Move dependencies to NuGet feed](https://trello.com/c/1ehBytwq/183-move-dependencies-to-nuget-feed)

appveyor.yml
```diff
@@ -7,16 +7,29 @@ assembly_info:
   assembly_file_version: "{version}"
   assembly_informational_version: "{version}"
 platform: x64
+environment:
+  nuget_user:
+    secure: kWtlS49bKmP0IFXmdvDHmxGI8mS0xQGrwiJoco9nYtA=
+  nuget_password:
+    secure: Pzvx2CQjxp8eCBP+nFI2eg==
+  appveyor_rdp_password:
+    secure: Pzvx2CQjxp8eCBP+nFI2eg==
```
FreexD: can we avoid storing passwords here?

appveyor.yml
```diff
@@ -7,16 +7,29 @@ assembly_info:
   assembly_file_version: "{version}"
   assembly_informational_version: "{version}"
 platform: x64
+environment:
+  nuget_user:
+    secure: kWtlS49bKmP0IFXmdvDHmxGI8mS0xQGrwiJoco9nYtA=
+  nuget_password:
+    secure: Pzvx2CQjxp8eCBP+nFI2eg==
```
FreexD: Can we avoid storing passwords here?

gmrukwa: I am not so sure - we have to pass them somehow, instead of hardcoding them in plaintext in scripts. These are password which allow read-only access to MATLAB NuGet package & to connect to build worker through RDP *during the build*.

appveyor.yml
gmrukwa: As above.

appveyor.yml
gmrukwa: [More info here](https://www.appveyor.com/docs/build-configuration/#secure-variables)

FreexD: I'm fine with that ;)

appveyor.yml
FreexD: as above

gmrukwa:
We will prune history from these files at some point, but this will be done after releasing new version.

PR #107, opened 2017-08-03
Coding standard enforcement

gmrukwa:
This is an enormous branch which changes just everything. But it should be fast, since these are mostly formatting stuffs.

* All code has been reformatted to obey our coding standard.
* ReSharper settings file has been added for solution. It will mark you all convention related issues and fix some of them for you (when you apply code cleanups).
* Warnings from linker due to GTest have been removed - it is necessary to install GMock *only* to resolve those warnings in further test projects.
* There appeared one warning from linker: `warning LNK4075: ignoring '/EDITANDCONTINUE' due to '/OPT:LBR' specification` (more [here](https://stackoverflow.com/questions/21965816/warning-lnk4075-when-a-c-cli-project-references-a-static-lib-project-with-zi)). It appears to be linker bug, since the mentioned `/OPT:LBR` option is explained to matter in the case of compilation for ARM.
* Ruleset for StyleCop has been added globally.
* Normalized .NET framework target version to 4.6.1

Some of StyleCop rules have been temporarily disabled, but some of them should be successively introduced soon:

* Copyright notice & file header checks
* Rules for comments of properties (indicating `get` or `get` and `set`), as well as return value tags there.
* Elements can now be placed in one line in the case of simple constructors, etc.
* Closing brackets do not have to be followed by an empty line.
* Type cast must not be followed by space.
* Braces are not required in arithmetic operations.


gmrukwa:
[Investigate StyleCop](https://trello.com/c/3NkU2rW9/190-investigate-stylecop)

gmrukwa:
This *lost* guideline appears due to proper configuration. Right now our code scores 7/10 guidelines, while previously it was 4/10, just due to wrong configuration. On the other hand, it shows some issues, that should in fact be addressed at some point.

PR #110, opened 2017-08-06
REST API packaging

gmrukwa:
Simplistic packaging of Web API project `Spectre` using *publish profiles* from Visual Studio.

**Note:** It is crucial to remember to not to put connection strings into those *publish profiles* in future. The purpose of those *publish profiles* in our project will be **only** to create a package during build. The real publishing process itself will be handled throught AppVeyor configuration & environment variables (i.e. to secure connection to deployment server and connection strings).


mg6:
@gmrukwa I would suggest using environment variables to pass GitHub deployment keys in `appveyor.yml`.

mg6:
@gmrukwa Nevermind, wasn’t aware those are already encrypted [secure variables](https://www.appveyor.com/docs/build-configuration/#secure-variables). All fine, then.

PR #123, opened 2017-08-24
Read spectra by coords

mg6:
- Allow querying for spectra using X, Y coordinates.
- Refactor existing code to throw HTTP 404 when queried for non-existent preparation/spectrum.

### DIFF
--- a/src/Spectre.Tests/Controllers/SpectrumControllerTest.cs
+++ b/src/Spectre.Tests/Controllers/SpectrumControllerTest.cs
@@ -18,6 +18,8 @@
 */
 
 using System.Linq;
+using System.Net;
+using System.Web.Http;
 using NUnit.Framework;
 using Spectre.Controllers;
 using Spectre.Models.Msi;
@@ -30,6 +32,8 @@ namespace Spectre.Tests.Controllers
     [TestFixture]
     internal class SpectrumControllerTest
     {
+        private const double Delta = 0.001;
+
         private SpectrumController _controller;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


mg6:
[Get Spectrum by location (API)](https://trello.com/c/LokVyLdh/221-get-spectrum-by-location-api)

src/Spectre.Tests/Controllers/SpectrumControllerTest.cs
```diff
@@ -42,21 +46,100 @@ public void SetUp()
...
+
+        /// <summary>
+        /// Tests getting non-existent spectrum.
+        /// </summary>
+        [Test]
+        public void TestThrows404ForInvalidSpectrum()
+        {
+            try
+            {
+                var spectrum = _controller.Get(id: 1, spectrumId: 999);
```
gmrukwa: Please name magic numbers.

src/Spectre/Controllers/SpectrumController.cs
```diff
@@ -39,13 +42,58 @@ public Spectrum Get(int id, int spectrumId)
...
+                    Id = spectrumId,
+                    Intensities = intensities,
+                    Mz = mz,
+                    X = coordinates.X,
+                    Y = coordinates.Y
+                };
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                throw new HttpResponseException(HttpStatusCode.NotFound);
```
gmrukwa: Good point.

src/Spectre.Tests/Controllers/SpectrumControllerTest.cs
mg6: Addressed this in 85d5476.

mg6:
@Daroslaw Agreed that we should provide messages for `Assert`s but, in my opinion, only when the assert method used and, consequently, its failure message themselves do not convey the intent clearly. Please consider the following examples:

```cs
// 1

var spectrum = _controller.Get(...);
Assert.Fail();  // this is not enough information

var spectrum = _controller.Get(...);
Assert.Fail("Should throw for invalid coords");  // now the assert’s purpose is clear
```

```cs
// 2

Assert.NotNull(spectrum); // clear intent, DRY
Assert.IsInstanceOf<Spectrum>(spectrum); // same as above

Assert.NotNull(spectrum, "Should return a non-null value");
    // redundant, already expressed with assert method
Assert.IsInstanceOf<Spectrum>(spectrum, "Should be of spectrum type");
    // same as above
```

```cs
// 3

catch (HttpResponseException e)
{
    Assert.AreEqual(actual: e.Response.StatusCode, expected: HttpStatusCode.NotFound);
        // comparison on enumerated values, clear intent
}

catch (HttpResponseException e)
{
    Assert.AreEqual(
        actual: e.Response.StatusCode,
        expected: HttpStatusCode.NotFound,
        message: "Should respond with proper HTTP code");  // feels redundant in my opinion
} 

catch (HttpResponseException e)
{
    Assert.AreEqual(
        actual: e.Response.StatusCode,
        expected: HttpStatusCode.NotFound,
        message: "Should respond with 404 Not Found code");
            // response code repeated twice, too specific
} 
```

I think the second case in `// 3` is disputable but will add it to address your concern.

gmrukwa:
@mg6 in my opinion, when have small tests (single assert), name of the test describes everything. Assertion messages are really important if we insert any greater number of asserts into single test, especially if they do not test the same functionality.

Daroslaw:
Alrighty then.

PR #124, opened 2017-08-24
Feature/dataset loader

Daroslaw:
Implemented DatasetLoader class.

Take note that tests that check if the code that actually returns some datasets are shit because BasicTextDataset does not support loading files from mocked file systems.


src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        private DataRootConfig _rootConfig;
+        private DatasetLoader _datasetLoader;
+        private MockFileSystem _mockFileSystem;
+
+        private readonly string _rootDir = @"C:\spectre_data";
+        private readonly string _localDir = @"\local";
+        private readonly string _remoteDir = @"\remote";
+
+        private readonly string _fileDir = TestContext.CurrentContext.TestDirectory
```
FreexD: Try using string interpolation for evaluating string expressions. It condenses the code and calls string format underneath, which makes it more localization safe than just adding strings.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        /// <summary>
+        /// Returns dataset basing directly on the name of the file.
+        /// </summary>
+        /// <param name="name">File name.</param>
+        /// <returns>Found dataset.</returns>
+        /// <exception cref="ArgumentException">Throws when the file is not found both locally and remotely.</exception>
+        public IDataset GetFromName(string name)
+        {
+            string fullPathLocal = _localRoot + "\\" + name;
```
FreexD: Try Path.Combine for that.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            if (!FileSystem.File.Exists(fullPathLocal))
+            {
+                string fullPathRemote = _remoteRoot + "\\" + name;
```
FreexD: Path.Combine

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+                if (!FileSystem.File.Exists(fullPathRemote))
+                {
+                    throw new ArgumentException("Dataset \"" + name + "\" not found neither locally nor remotely.");
```
FreexD: Use nameof operator for throwing ArgumentException - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/nameof .

FreexD: Try using string interpolation for evaluating string expressions. It condenses the code and calls string format underneath, which makes it more localization safe than just adding strings.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
FreexD: Or you could just use Path.Combine here :)

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,115 @@
...
+                                           + "\\..\\..\\..\\..\\..\\test_files";
+        [SetUp]
+        public void SetUp()
+        {
+            var localDirFull = _rootDir + _localDir;
```
FreexD: Try Path.Combine for that.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+        #endregion
+
+        #region Loaders
+
+        /// <summary>
+        /// Returns dataset basing directly on the name of the file.
+        /// </summary>
+        /// <param name="name">File name.</param>
+        /// <returns>Found dataset.</returns>
+        /// <exception cref="ArgumentException">Throws when the file is not found both locally and remotely.</exception>
```
FreexD: This method throws IOException as well.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+                }
+
+                FileSystem.File.Copy(fullPathRemote, fullPathLocal);
+            }
+
+            IDataset result;
+
+            try
+            {
+                result = new BasicTextDataset(fullPathLocal);
```
FreexD: Why not return here?

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,115 @@
...
+            _mockFileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                { localDirFull, new MockDirectoryData() },
+                { remoteDirFull, new MockDirectoryData() },
+                { localDirFull + @"\local_correct.txt", new MockFileData(correctDataset) },
+                { remoteDirFull + @"\remote_correct.txt", new MockFileData(correctDataset) },
+                { localDirFull + @"\local_incorrect.txt", new MockFileData(textContents: "incorrect_data") }
+            });
+
+            var kernel = new StandardKernel();
```
FreexD: Nice touch ;)

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,137 @@
...
+            _remoteRoot = dataRootConfig.RemotePath;
+        }
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Handle to file system.
+        /// </summary>
+        [Inject]
```
FreexD: Nice touch ;)

src/Spectre.Service/Loaders/DatasetLoader.cs
Daroslaw: Done

src/Spectre.Service/Loaders/DatasetLoader.cs
Daroslaw: thx

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,115 @@
...
+            _rootConfig = new DataRootConfig(localDirFull, remoteDirFull);
+
+            _datasetLoader = kernel.Get<DatasetLoader>(new ConstructorArgument(name: "dataRootConfig", value: _rootConfig));
+        }
+
+        [Test]
+        public void ReturnsFromCorrectNameLocal()
+        {
+            Assert.Throws<IOException>(code: () => _datasetLoader.GetFromName(name: "local_correct.txt"), 
+                                                message: "Loader did not manage to load local file.");
```
gmrukwa: Indents in this file are broken. Please use spaces instead of tabs - they should always work.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            _rootConfig = new DataRootConfig(localDirFull, remoteDirFull);
+
+            _datasetLoader = kernel.Get<DatasetLoader>(new ConstructorArgument(name: "dataRootConfig", value: _rootConfig));
+        }
+
+        [Test]
+        public void ReturnsFromCorrectNameLocal()
+        {
+            Assert.Throws<IOException>(code: () => _datasetLoader.GetFromName(name: "local_correct.txt"),
```
gmrukwa: I think that we should read `IFileSystem` from `Ninject` during construction of `BasicTextDataset`. This way we will be able to test this properly for some mock data.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,134 @@
...
+        public IDataset GetFromName(string name)
+        {
+            string fullPathLocal = Path.Combine(_localRoot, name);
+
+            if (!FileSystem.File.Exists(fullPathLocal))
+            {
+                string fullPathRemote = Path.Combine(_remoteRoot, name);
+                if (!FileSystem.File.Exists(fullPathRemote))
+                {
+                    throw new ArgumentException($"Dataset '{name}' not found neither locally nor remotely.", nameof(name));
```
gmrukwa: I think there should be dedicated exception for this purpose.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            try
+            {
+                return new BasicTextDataset(fullPathLocal);
+            }
+            catch (IOException)
+            {
+                FileSystem.File.Delete(fullPathLocal);
+                throw;
```
gmrukwa: Here should also be *another* dedicated exception.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
Daroslaw: B-but those are spaces

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            _rootConfig = new DataRootConfig(localDirFull, remoteDirFull);
+
+            _datasetLoader = kernel.Get<DatasetLoader>(new ConstructorArgument(name: "dataRootConfig", value: _rootConfig));
+        }
+
+        [Test]
+        public void ReturnsFromCorrectNameLocal()
+        {
+            Assert.Throws<DatasetLoadException>(code: () => _datasetLoader.GetFromName(name: "local_correct.txt"),
```
gmrukwa: I am unsure what is happening here.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,134 @@
...
+        public IDataset GetFromName(string name)
+        {
+            string fullPathLocal = Path.Combine(_localRoot, name);
+
+            if (!FileSystem.File.Exists(fullPathLocal))
+            {
+                string fullPathRemote = Path.Combine(_remoteRoot, name);
+                if (!FileSystem.File.Exists(fullPathRemote))
+                {
+                    throw new FileNotFoundException("File not found neither locally nor remotely.", name);
```
gmrukwa: What about `DatasetNotFoundException`?

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            try
+            {
+                return new BasicTextDataset(fullPathLocal);
+            }
+            catch (IOException e)
+            {
+                FileSystem.File.Delete(fullPathLocal);
+                throw new DatasetLoadException($"Failed to load dataset from file '{name}'.", e);
```
gmrukwa: It fails to load when it's incorrectly formatted, doesn't it? Maybe something like `DatasetFormatException`? Or we do cover any other cases?

src/Spectre.Service/Loaders/DatasetLoader.cs
Daroslaw: For this purpose is such specification neccessary?

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
Daroslaw: This test requires BasicTextDataset to handle its IO from the abstractions we've discussed. For now the loader is able to read the mocked file system but when it passes a mocked path, the exception is thrown.

src/Spectre.Service/Loaders/DatasetLoader.cs
Daroslaw: I thought there could be some other cases that may be generalized with such exception name. Maybe I'm mistaken. I will change that.

src/Spectre.Service/Loaders/DatasetLoader.cs
gmrukwa: What I have in mind are logs, which we will enable in future. Usually, the exception name is saved, sometimes with message, call stack, etc. The more precise we are here, the better we will be able to debug from the logs. What will tell us `FileNotFoundException` for 2-file `imzml` file? Is this `.imzml` file missing or `.ibd`? One of these will be our first guess, while the case is that a dataset does not exist at all.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
gmrukwa: Is this a lot of work to migrate `Dataset` to `IFileSystem`?

gmrukwa:
[Prepare simple loader of the datasets](https://trello.com/c/uT44i2lT/173-prepare-simple-loader-of-the-datasets)

src/Spectre.Data/Datasets/BasicTextDataset.cs
```diff
@@ -110,6 +114,15 @@ public int SpectrumCount
         }
 
         /// <summary>
+        /// Handle to file system.
+        /// </summary>
+        public IFileSystem FileSystem { private get; set; }
```
FreexD: Or [this](https://stackoverflow.com/a/26797938) way. Be careful however. Make sure we dont have to destroy provided services manually.

src/Spectre.Dependencies/FileSystemDependency.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.IO.Abstractions;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Ninject;
+
+namespace Spectre.Dependencies
+{
+    public static class FileSystemDependency
```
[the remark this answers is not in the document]
FreexD: This is how I think it could work: [link](https://ibb.co/et3DPF).

src/Spectre.Dependencies/FileSystemDependency.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        private static IKernel _kernel;
+
+        public static void Initialize(IFileSystem fileSystem)
```
FreexD: In initialize we should set a proper dependency resolver.

src/Spectre.Data/Datasets/BasicTextDataset.cs
FreexD: Or using [Inject] annotation for property.

src/Spectre.Data/Spectre.Data.csproj
```diff
@@ -34,8 +34,32 @@
     <DocumentationFile>bin\x64\Release\Spectre.Data.xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Web.Infrastructure, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
FreexD: Could we find a way get rid of at least some of these dependencies?

src/Spectre.Service/Spectre.Service.csproj
```diff
@@ -35,14 +35,53 @@
     <DocumentationFile>bin\x64\Release\Spectre.Service.xml</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Web.Infrastructure, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
FreexD: Same here.

src/Spectre.Data/Spectre.Data.csproj
FreexD: Honestly I don't know what to do about that. I'll ask colleagues and stack overflow.

src/Spectre.Service/Dependencies/MockFilesSystemModule.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.IO.Abstractions.TestingHelpers;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Ninject.Modules;
+
+namespace Spectre.Service.Dependencies
+{
+    /// <summary>
+    /// To do
```
FreexD: Fix all documentation comments.

src/Spectre.Data/Spectre.Data.csproj
Daroslaw: Well, that what has been bothering me. Lots of MVC-related dependencies but I was like "m'kay, if we need this in order to do it properly and in future-safe way, so be it".

src/Spectre.Service/Dependencies/MockFilesSystemModule.cs
Daroslaw: Documentation will be added if some final solution is found,

FreexD:
We have decided to solve DependencyResolver MVC dependency problem by exposing a singleton ninject kernel  and adding a possibility to load necessary bindings dependending on their application (tests, webApi, etc.), without the need to use DependencyResolver for the sake of this task.

src/Spectre.Data/Datasets/BasicTextDataset.cs
gmrukwa: This one shouldn't be `readonly`?

src/Spectre.Dependencies/DependencyResolver.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        /// <param name="newModule">Module to load.</param>
+        public static void AddModule(INinjectModule newModule) => Kernel.Load(newModule);
+
+        /// <summary>
+        /// Method for getting a certain service according to its type.
+        /// </summary>
+        /// <param name="serviceType">Type of desired service.</param>
+        /// <returns></returns>
+        public static object GetService(Type serviceType) => Kernel.TryGet(serviceType);
```
gmrukwa: This would be better as generic, with changing result type:
`public static T GetService<T>() => Kernel.TryGet(typeof(T)) as T;`
or something like this. This way we do not have to try to cast each time.

src/Spectre.Dependencies/Modules/MockModule.cs
```diff
@@ -0,0 +1,45 @@
+﻿/*
+ * DefaultModule.cs
```
gmrukwa: Old name here.

src/Spectre.Dependencies/Spectre.Dependencies.csproj
```diff
@@ -0,0 +1,62 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
```
gmrukwa: We have `AnyCPU` platform here.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,109 @@
...
+        private DatasetLoader _datasetLoader;
+        private MockFileSystem _mockFileSystem;
+
+        private readonly string _rootDir = @"C:\spectre_data";
+        private readonly string _localDir = "local";
+        private readonly string _remoteDir = "remote";
+
+        private readonly string _fileDir = $"{TestContext.CurrentContext.TestDirectory} + "
+                                           + @"\..\..\..\..\..\test_files";
+        [OneTimeSetUp]
```
gmrukwa: Is this sufficient to be `OneTime`?

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        [Test]
+        public void DeletesIncorrectFilesFromLocal()
+        {
+            try
+            {
+                _datasetLoader.GetFromName(name: "remote_incorrect.txt");
+            }
+            catch (Exception)
```
gmrukwa: If it is ignored, we should listen for specific type of exception to not to hide errors in tests.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        [Test]
+        public void ThrowsOnIncorrectFileContents()
+        {
+            Assert.Throws<DatasetFormatException>(code: () => _datasetLoader.GetFromName(name: "local_incorrect.txt"),
+                message: "Loader did not manage to load remote file.");
+        }
+
+        [Test]
+        public void DeletesIncorrectFilesFromLocal()
```
gmrukwa: Nice functionality!

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
```diff
@@ -0,0 +1,109 @@
...
+            try
+            {
+                _datasetLoader.GetFromName(name: "remote_incorrect.txt");
+            }
+            catch (Exception)
+            {
+                // ignored
+            }
+            var result = _mockFileSystem.AllFiles.FirstOrDefault(predicate: file => file.Contains(value: Path.Combine(_localDir, "remote_incorrect.txt")));
+            Assert.IsNull(result, message: "Loader leaves copies of incorrect files in local directory.");
```
gmrukwa: Does remote remain untouched? It is good, but I just want to confirm.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,136 @@
...
+            FileSystem = DependencyResolver.GetService(typeof(IFileSystem)) as IFileSystem;
+        }
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Handle to file system.
+        /// </summary>
+        public IFileSystem FileSystem { private get; set; }
```
gmrukwa: Shouldn't it be `readonly`? Why `set` is `public`? Especially when `get` is `private`? It is assumed to be an antipattern. I thought we are switching this only in the tests, before any call to kernel is done.

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            try
+            {
+                return new BasicTextDataset(fullPathLocal);
+            }
+            catch (IOException e)
+            {
+                FileSystem.File.Delete(fullPathLocal);
+                throw new DatasetFormatException($"Failed to load dataset from file '{name}'.", e);
```
gmrukwa: This should be inside `BasicTextDataset` constructor (I suppose).

src/Spectre.Service/Loaders/DatasetLoader.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+
+        /*
+        //  TODO: @dkuchta: complete the methods below when database access is available
```
gmrukwa: Add it on Trello, either as a task, or extend the database-related task.

src/Spectre.Service/Loaders/DatasetNotFoundException.cs
```diff
@@ -0,0 +1,52 @@
...
+        public DatasetNotFoundException(string message, string fileName) : base(message, fileName)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DatasetNotFoundException"/> class.
+        /// </summary>
+        /// <param name="message">The exception message.</param>
+        /// <param name="innerException">The original exception thrown during validation.</param>
+        public DatasetNotFoundException(string message, Exception innerException) : base(message, innerException)
```
gmrukwa: File name is missing. Isn't there an option to enforce passing it here? This may be relevant from the point of view of debugging.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
Daroslaw: It is. Another solution that I've thought of is to make OneTime module addition to DependencyResolver and the rest to ordinary SetUp. But this works fine.

src/Spectre.Service.Tests/Loaders/DatasetLoaderTest.cs
Daroslaw: Yeah, the file persists on the remote side.

src/Spectre.Service/Loaders/DatasetLoader.cs
Daroslaw: BasicTextDataset throws IOExceptions in this case. They are quite informative imho.

gmrukwa: Let it be then, at least for now.

src/Spectre.Service/Loaders/DatasetLoader.cs
Daroslaw: Done

PR #127, opened 2017-08-28
Fix native tests not failing properly

mg6:
- Fix variable increment as per https://stackoverflow.com/a/25632806.
- Simplify code a bit.

### DIFF
--- a/scripts/test_native_code.bat
+++ b/scripts/test_native_code.bat
@@ -9,10 +9,7 @@ set failures=0
 
 FOR /F %%i IN (__tmp_gtest.txt) DO (
 	echo %%i
-	%%i --gtest_output="xml:%%i.xml"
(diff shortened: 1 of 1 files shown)


scripts/test_native_code.bat
```diff
@@ -9,11 +9,8 @@ set failures=0
 
 FOR /F %%i IN (__tmp_gtest.txt) DO (
 	echo %%i
-	%%i --gtest_output="xml:%%i.xml"
+	%%i --gtest_output="xml:%%i.xml" || set /A failures=failures+1
 	powershell C:\projects\spectre\scripts\Upload-TestResult.ps1 -fileName %%i.xml
-	IF %ERRORLEVEL% NEQ 0 (
-		set /A failures=%failures%+1
```
gmrukwa: This one was from SO as well :D

PR #131, opened 2017-09-17
divik level

Sebaxp95:


src/Spectre.AngularClient/src/app/divik-results/divik/divik.component.html
```diff
@@ -3,9 +3,19 @@
   {{configDescription}}
 </div>
 
+<md-form-field *ngIf="divikConfig">
+  <input mdInput
+         type='number'
+         min="0"
+         max={{divikConfig.Level}}
+         (change)="changeLevel($event.target.value)"/>
+  <md-hint align="center">Max value: divikConfig.Level</md-hint>
```
gmrukwa: This is not necessarily true. If the experiment has been ran with `useLevels: false`, backend should estimate tree depth and exhibit it to the frontend.

src/Spectre.AngularClient/src/app/divik-results/divik/divik.component.ts
```diff
@@ -65,6 +67,15 @@ export class DivikComponent implements OnInit {
     return description;
   }
 
+  changeLevel(value) {
+    console.log('change ' + value);
+    if (value > 0 && value <= this.divikConfig.Level) {
+      this.divikService
+        .get(1, 1, value)
```
gmrukwa: We should use known `preparationId`, shouldn't we? It has been passed through the router.

src/Spectre.AngularClient/src/app/divik-results/divik/divik.component.html
Sebaxp95: So if we haven't had this information yet it could be another task

gmrukwa:
[Enable DiviK depth selection](https://trello.com/c/g5Q58WR8/211-enable-divik-depth-selection)

PR #142, opened 2017-10-24
Data validator application

gmrukwa:
I've prepared a converter of MATLAB data files (used to provide data to DiviK) into our text format. No black box tests are provided, since it's just a runnable script and I would like to not to dig into mat-file internal binary format to mock it. White box testing (assertions) are supplied in the respective parsing functions. Converter works both for v7.3 and lower version files.

To test it more reliably, I've created a micro app just for validation of text data file format.


gmrukwa:
[Make python .mat converter](https://trello.com/c/ZIUpviDj/238-make-python-mat-converter)

src/Python/divik2spectre.py
```diff
@@ -0,0 +1,149 @@
...
+        assert mz.size == data.shape[1]
+    return DivikDataSerializer(data, xy, mz, region, region_names)
+
+
+def main():
+    args = parse_args()
+    try:
+        data = read_matfile(args.source)
+    except NotImplementedError:
+        data = read_hdf5(args.source)
```
mg6: Just wondering, where can `read_matfile` raise a `NotImplementedError`?

gmrukwa: It raises when we're trying to read a `-v7.3` file with `scipy` reader - it does not implement `hdf5` files. This way we have automated version detection instead of forcing user to pass this information.

mg6: Sure thing. Thanks for clarification!

docs/Spectre.DataValidatorApp.md
```diff
@@ -0,0 +1,71 @@
+---
```
FreexD: divik2spectre.py is not mentioned here. What if users would like to use it? Is that script just for us?

src/Python/divik2spectre.py
```diff
@@ -0,0 +1,149 @@
...
+def read_matfile(path: str) -> DivikDataSerializer:
+    """Parses v7.0 >= mat-file with DiviK data into serializer"""
+    matfile = scio.loadmat(path)
+    region_names = [name_table[0] for name_table in matfile['region_names'][0]]
+    data = matfile['data']
+    xy = matfile['xy']
+    assert data.shape[0] == xy.shape[0]
+    assert xy.shape[1] == 2
+    region = matfile['region']
+    assert region.size == data.shape[0]
```
FreexD: The whole script looks solid. No tests though :P

src/Python/divik2spectre.py
```diff
@@ -0,0 +1,149 @@
...
+import argparse as agp
+import h5py
+import numpy as np
+import scipy.io as scio
+from tqdm import tqdm
+from typing import List
+
+
+def parse_args():
+    """Parses command line arguments, requiring source and destination paths"""
```
FreexD: If it's not a problem, please specify converter input format somewhere in the docs of this script or the project (or just provide a link to description).

src/Python/divik2spectre.py
FreexD: n1

src/Spectre.DataValidatorApp/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            if ((args.Length != 1) || ((args.Length == 2) && (args[1] != "--nonblocking")))
+            {
+                Console.Write(value: $"Usage: {AppDomain.CurrentDomain.FriendlyName} path_to_data_file [--nonblocking]");
+            }
+            else
+            {
+                try
+                {
+                    new BasicTextDataset(textFilePath: args[0]);
```
FreexD: n1

src/Python/divik2spectre.py
```diff
@@ -0,0 +1,149 @@
...
+def _serialize_entry(spectrum, location, region):
+    """Serializes single spectrum with metadata
+
+    :param spectrum: single spectrum values
+    :param location: 2D coordinates of the spectrum
+    :param region: assignment of spectrum to a Region-of-Interest
+    :return: text file entry
+    """
+    return '\n'.join([
+        ' '.join(map(str, [*location, CONST_Z_COORDINATE, region])),
```
FreexD: n1

docs/Spectre.DataValidatorApp.md
gmrukwa: This script is intended for our internal purposes *only*.

src/Python/divik2spectre.py
gmrukwa: Added on the top of the script.

src/Python/divik2spectre.py
gmrukwa: I know. This is due to several reasons:
1) There's only one functionality, that could be unit-tested properly: serializer. Rest of the script is I/O or file format related (I don't know the internals of MAT-file in a way I could mock it), so it would require hacky testing.
2) White-box testing was applied through `asserts` wherever I've seen such functionality that is untestable from the outside (through black-box testing).
3) I've created a kind of integration test - the validator application, which for me served *only* for validation of this script. I just thought it may be useful in the long run for the users also.

FreexD: SGTM.

PR #144, opened 2017-10-29
Fixup/random splitter discards seed

gmrukwa:
Something strange happened in the code and random seed was never used at all.


gmrukwa:
[Fix RandomSplitter](https://trello.com/c/0LKlhllW/281-fix-randomsplitter)

PR #146, opened 2017-11-06
Feature/dataset saver

Daroslaw:


src/Spectre.Service.Tests/Savers/DatasetSaverTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        [Test]
+        public void SavesFromMemory()
+        {
+            double[] mz = { 1.0, 2.0, 3.0 };
+            double[,] data = { { 1, 2.1, 3.2 }, { 4, 5.1, 6.2 }, { 7, 8.1, 9.2 } };
+            IDataset testDataset = new BasicTextDataset(mz, data, null);
+            _datasetSaver.SaveFromMemory(testDataset, "testDataset");
+            Assert.IsNotEmpty(collection: _mockFileSystem.Directory.GetFiles(Path.Combine(_rootDir, _remoteDir),
```
gmrukwa: Do we only want to save to remote here? Are we absolutely skipping local?

src/Spectre.Service/Savers/DatasetSaver.cs
```diff
@@ -0,0 +1,105 @@
...
+
+using System.IO;
+using System.IO.Abstractions;
+using System.Linq;
+using Spectre.Data.Datasets;
+using Spectre.Dependencies;
+using Spectre.Service.Configuration;
+using Spectre.Service.Loaders;
+
+namespace Spectre.Service.Savers
```
gmrukwa: Maybe it could be together with `DatasetLoader` inside something like `Spectre.Service.Io`? For a while we won't support another savers.

src/Spectre.Service/Savers/DatasetSaver.cs
```diff
@@ -0,0 +1,105 @@
...
+            FileSystem = DependencyResolver.GetService<IFileSystem>();
+        }
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Handle to file system.
+        /// </summary>
+        private IFileSystem FileSystem { get; }
```
gmrukwa: Can't it be `readonly`?

src/Spectre.Service/Savers/DatasetSaver.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        /// <summary>
+        /// Method for saving given dataset from application's memory into remote root.
+        /// </summary>
+        /// <param name="dataset">Dataset to be saved.</param>
+        /// <param name="name">User-friendly name given to the dataset.</param>
+        public void SaveFromMemory(IDataset dataset, string name)
+        {
+            string extension = ".txt"; // TODO @dkuchta: extension recognition
```
gmrukwa: I don't think we are able to make *recognition* here - this will be arbitrarily selected by us :)

src/Spectre.Service/Savers/DatasetSaver.cs
Daroslaw: Fair point

src/Spectre.Service/Savers/DatasetSaver.cs
Daroslaw: Without 'set' the property is implicitly read-only.

src/Spectre.Service.Tests/Savers/DatasetSaverTest.cs
Daroslaw: Should we also save to local?

gmrukwa:
Continued as: https://github.com/spectre-team/platform-components/pull/5

gmrukwa:
[Dataset store](https://trello.com/c/eak9PgmB/144-dataset-store)

gmrukwa:
[Prepare simple saver of the datasets](https://trello.com/c/OkSHuM5h/259-prepare-simple-saver-of-the-datasets)

PR #147, opened 2017-11-08
MATLAB result loader

gmrukwa:
Goal of the task was to provide a way to convert existing results of analyses (enclosed in MAT-files) into a format we support - JSON.


gmrukwa:
[Converter of current MATLAB DiviK results into JSON](https://trello.com/c/SMU47NM9/291-converter-of-current-matlab-divik-results-into-json)

src/Spectre.Algorithms.Tests/Io/DivikResultLoaderTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            Assert.DoesNotThrow(
+                code: () =>
+                {
+                    using (var loader = new DivikResultLoader())
+                    { }
+                });
+        }
+
+        [Test]
+        public void LoadsTree()
```
Daroslaw: This could be separated into several tests.

PR #151, opened 2017-11-12
Split libPlatform

gmrukwa:
This is the subset of changes I am rushing in the run of molecular signature estimation framework. I thought it'll be useful for you to see it as a self-contained subpart.

Here I've splitted `libPlatform` to separate the functionalities gathered there. This way, we'll not push too much responsibilities into this project.

Right now we will have two projects:

* `libFunctional` - here are all functional-alike extensions for C++ we're developing
* `libStatistics` - here I will push some statistical indexes, etc., just to gather them together


gmrukwa:
[Refactor libPlatform](https://trello.com/c/m2PRlhce/309-refactor-libplatform)

PR #155, opened 2017-11-13
Split libPlatform

gmrukwa:
This is the subset of changes I am rushing in the run of molecular signature estimation framework. I thought it'll be useful for you to see it as a self-contained subpart.

Here I've splitted `libPlatform` to separate the functionalities gathered there. This way, we'll not push too much responsibilities into this project.

Right now we will have two projects:

* `libFunctional` - here are all functional-alike extensions for C++ we're developing
* `libStatistics` - here I will push some statistical indexes, etc., just to gather them together

PR #156, opened 2017-11-14
Signature estimation framework

gmrukwa:
Proceed with this PR only if #155 is already merged, to see proper (smaller) subset of changes.

`libStatistics` have been supplied with simple framework for estimation of feature differentiation strength. To accomplish this, some minor (but distributed) changes had to be made also in other projects:

1. `Empty` temporary fix for `IDataset` not supporting disabling metadata through setting it to `void`. It has been moved from `libClassifier` to the root of the problem - `libDataset`.
2. `OutOfRangeException` appeared to be non-resolved by compiler when trying to use `Dataset<T1, T2, T3>`. Proper `#include` had to be made, and explicit references to `libException::OutOfRangeException`.


gmrukwa:
[Molecular signature analysis framework (C++)](https://trello.com/c/lohfRqea/311-molecular-signature-analysis-framework-c)

src/Spectre.libStatistics/StatisticalIndex.cpp
```diff
@@ -0,0 +1,28 @@
+/*
+* OrderableIndex.cpp
+* ENTER CLASS DESCRIPTION HERE.
```
Daroslaw: Yep

src/Spectre.libClassifier.Tests/ObservationExtractorTest.cpp
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 */
 
 #include <gtest/gtest.h>
+#include <memory>
```
Daroslaw: Simply a reference to memory STL solves ambiguity for `OutOfRangeException`? I thought this makes an opposite effect. 😨

PR #157, opened 2017-11-15
Cohen's D effect size

gmrukwa:
Before proceeding with this pull request, make sure #156 is already closed and merged, to see only limited subset of changes. Otherwise you will see a lot of stuff unrelated to current changeset.

I've introduced few new exceptions. Some of them are general ones and placed in `libException`. In order to implement Cohen's D effect size, variance for two samples needs to be computed - it's implemented here as well.


gmrukwa:
[Add Cohen's D effect size feature strength measure](https://trello.com/c/VeLtk57N/312-add-cohens-d-effect-size-feature-strength-measure)

src/Spectre.libStatistics/StatisticalIndex.cpp
```diff
@@ -0,0 +1,28 @@
+/*
+* OrderableIndex.cpp
+* ENTER CLASS DESCRIPTION HERE.
```
Daroslaw: This PR is so meta, even the source gives you useful hints

gmrukwa: I've created file templates for VS, so these things became easy to catch :P

src/Spectre.libException/EmptyDatasetException.h
```diff
@@ -0,0 +1,37 @@
...
+
+#pragma once
+#include "Spectre.libException/EmptyArgumentException.h"
+
+namespace Spectre::libException
+{
+/// <summary>
+/// Thrown when input dataset was empty.
+/// </summary>
+class EmptyDatasetException final: public EmptyArgumentException
```
mg6: Formatting, `final:` → `final :`, as in `ReachedUnreachableCodeException`.

src/Spectre.libStatistics/CohenEffectSize.cpp
```diff
@@ -0,0 +1,74 @@
...
+    const auto std = sqrt(Variance(first, second));
+    const auto DCohen = std::abs(Mean(first) - Mean(second)) / sqrt(Variance(first, second));
+    for (auto thresholdNumber = thresholds.size() - 1; thresholdNumber >= 0; --thresholdNumber)
+    {
+        if (DCohen >= thresholds[thresholdNumber])
+        {
+            return StatisticalIndex(DCohen, static_cast<unsigned>(thresholdNumber), interpretations[thresholdNumber]);
+        }
+    }
+    throw libException::ReachedUnreachableCodeException();
```
mg6: Suppose there are 2 paths which can yield a `ReachedUnreachableCodeException`. How the user is supposed to know which one failed? I would suggest either:
a) introducing a derived exception of `ReachedUnreachableCodeException` or
b) extending `ReachedUnreachableCodeException`’s constructor to accept `const std::string& message`
to exactly express the situation that none of the thresholds matched.

src/Spectre.libException/EmptyDatasetException.h
gmrukwa: Accepted.

src/Spectre.libStatistics/CohenEffectSize.cpp
gmrukwa: I've assumed the user has stack trace, hasn't he?

mg6: Well, yes and no. Why not both? There user can have the information *what* happened exactly, in addition to *where* it was (in case of stack trace).

gmrukwa: Reasonable idea.

gmrukwa: Just found this: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0627r0.pdf

Maybe that's the way to go?

mg6: Okay, that seems reasonable.

PR #163, opened 2018-01-02
Prune repository

gmrukwa:
Repository got pruned out of the code migrated to separate repos. It was all engineer Nachamov method, so it would be good, if you checked, whether your code is where it is expected to be.

I've had to remove `AlgorithmsCli`. It was one of our missed experiments we had to do at some point, that should be a precious knowledge for future. Its functionality will be in the worker repo, when the time comes.

Count of removed lines is mostly due to removed redundant test files. I do not expect anyone to read this line-by-line. Just open the solution and take a look at the changes. Another reasonable point to check is `DivikResult` and `DivikResultTests` - these had to be updated after removal of `AlgorithmsCli`.


gmrukwa:
[Extract common .NET modules to NuGets](https://trello.com/c/5SSNPeaF/333-extract-common-net-modules-to-nugets)

gmrukwa:
[Prune main repository](https://trello.com/c/AGhmH4hj/326-prune-main-repository)

PR #165, opened 2018-01-29
Add upload component

Sebaxp95:


Sebaxp95:
[Upload-frontend](https://trello.com/c/AsVALw6H/376-upload-frontend)

src/Spectre.AngularClient/src/app/upload/upload.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Http, BaseRequestOptions, Response, ResponseOptions } from '@angular/http';
+import { MockBackend} from '@angular/http/testing';
+
+import { RouterTestingModule } from '@angular/router/testing';
+import { UploadComponent } from "./upload.component";
+import { UploadService } from "./shared/upload.service";
+import {FormsModule} from "@angular/forms";
+
+
+describe('UploadComponent', () => {
```
gmrukwa: What about tests against enabling & disabling the button in the case of valid / invalid data?

src/Spectre.AngularClient/src/app/upload/shared/upload.service.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+          UploadService
+      ]
+    });
+  });
+
+  it('should inject UploadService', inject([UploadService], (service: UploadService) => {
+    expect(service).toBeTruthy();
+  }));
+
+  it('should return 200', inject([UploadService, MockBackend],
```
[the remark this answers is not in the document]
Sebaxp95: I mocked backend so I think it is like return from real server. When do you want return 200 after schedule or after upload?

src/Spectre.AngularClient/src/app/upload/shared/upload.service.ts
```diff
@@ -0,0 +1,53 @@
...
+@Injectable()
+export class UploadService extends Service {
+
+  constructor(private http: Http) {
+    super();
+  }
+
+  uploadData(datasetLink: string, datasetName): Observable<string> {
+    return this.http
+      .post(`${this.getBaseUrl()}/download/`, {headers: this.getHeaders(), url: datasetLink, datasetName: datasetName})
```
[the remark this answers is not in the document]
Sebaxp95: I can move all links to properties file

src/Spectre.AngularClient/src/app/upload/shared/upload.service.spec.ts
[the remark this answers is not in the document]
gmrukwa: What I mean is that `'should return 200'` suggests, that we are testing, whether mock works as intended. This misses the point of unit testing. We should test, what's the reaction of the client to status code `200`.

src/Spectre.AngularClient/src/app/upload/shared/upload.service.ts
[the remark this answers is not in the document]
gmrukwa: I think it's a good moment for this. I am starting to reorganize this repo to leave only the client, while all of the remaining components will be available through docker. This way we will have to be able to manipulate all of the API urls easily.

src/app/upload/shared/upload.service.ts
```diff
@@ -0,0 +1,53 @@
...
+@Injectable()
+export class UploadService extends Service {
+
+  constructor(private http: Http) {
+    super();
+  }
+
+  uploadData(datasetLink: string, datasetName): Observable<boolean> {
+    return this.http
+      .post(`${this.getBaseUrl()}/download/`, {headers: this.getHeaders(), url: datasetLink, datasetName: datasetName})
```
gmrukwa: Download service is separate machine serving the transfers. Extract its url to environment configs.

src/app/upload/shared/upload.service.ts
```diff
@@ -0,0 +1,53 @@
...
+      .map(r => {
+        if (r.status === 200) {
+          return true;
```
gmrukwa: Do we have to return anything? I don't know what it looks like in Angular, but maybe we don't have to?

package.json
```diff
@@ -65,6 +65,7 @@
     "@angular/platform-browser-dynamic": "^5.0.0-rc.1",
     "@angular/platform-server": "^5.0.0-rc.1",
     "@angular/router": "^5.0.0-rc.1",
+    "@types/sprintf": "^0.1.0",
```
gmrukwa: What does it do for us?

src/app/divik-results/shared/divik.service.ts
```diff
@@ -41,13 +43,13 @@ export class DivikService extends Service {
   }
 
   get(preparationId: number, divikId: number, level: number): Observable<Heatmap> {
-    const queryUrl = `${this.getBaseUrl()}/divikResult/${preparationId}?divikId=${divikId}&level=${level}`;
+    const queryUrl = this.getBaseUrl() + sprintf(apiUrl.divikResultUrl, preparationId, divikId, level);
```
gmrukwa: Do you specify the format somehow?

src/app/upload/shared/upload.service.ts
```diff
@@ -0,0 +1,49 @@
...
+@Injectable()
+export class UploadService extends Service {
+
+  constructor(private http: Http) {
+    super();
+  }
+
+  uploadData(datasetLink: string, datasetName): Observable<any> {
+    return this.http
+      .post(this.getBaseUrl() + apiUrl.uploadUrl, {headers: this.getHeaders(), url: datasetLink, datasetName: datasetName})
```
gmrukwa: `apiUrl` is not necessarily placed on `BaseUrl`. Right now we have separate ports for this, but this usually means separate hosts in fact.

src/app/upload/shared/upload.service.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+          },
+          UploadService
+      ]
+    });
+  });
+
+  it('should inject UploadService', inject([UploadService], (service: UploadService) => {
+    expect(service).toBeTruthy();
+  }));
+
```
gmrukwa: Test for a case without failure is missing. We had it in previous version, why did it get deleted? We'd just like to check, whether it passes without an error, right?

src/environments/apiUrl.ts
```diff
@@ -0,0 +1,11 @@
+export const apiUrl = {
+  spectrumByIdUrl: '/spectrum/%d?spectrumId=%d',
```
gmrukwa: API for visualization is on port `2001` since now.

src/environments/apiUrl.ts
```diff
@@ -0,0 +1,11 @@
...
+  spectrumByCoordUrl: '/spectrum/%d?x=%d&y=%d',
+  preparationsUrl: '/preparations',
+  preparationUrl: '/preparations/%s',
+  heatmapUrl: '/heatmap/%s?channelId=%s&flag=false',
+  divikResultUrl: '/divikResult/%s?divikId=#d&level=#s',
+  divikConfigUrl: '/divikResult/%s?divikId=%s',
+  uploadUrl: '/download'
```
gmrukwa: This API is on port `2002` since now: `my_host:2002/api/download`.

package.json
Sebaxp95: It is small library with function sprintf. It replace flags in string e.g %s with value from arguments list.

gmrukwa: Do we need additional dependency for this? I'd like at least to *try avoiding* adding too much dependencies...

src/app/divik-results/shared/divik.service.ts
```diff
@@ -41,13 +43,13 @@ export class DivikService extends Service {
   }
 
   get(preparationId: number, divikId: number, level: number): Observable<Heatmap> {
-    const queryUrl = `${this.getBaseUrl()}/divikResult/${preparationId}?divikId=${divikId}&level=${level}`;
+    const queryUrl = this.getBasePreparationUrl() + sprintf(apiUrl.divikResultUrl, preparationId, divikId, level);
```
gmrukwa: The ultimate goal of extraction of these urls to single place is to have single point for all changes involving migration of functionalities between workers / other hosts providing API. In current form, if only DiviK gets detached from visualization API, we still have to dig into the code - it's maintainability is low.

We can even agree on creating `getDivikUrl` which returns *the same value* as `getBasePreparationUrl` (for now!), but is maintainable from one central point. When there comes a need to change the machine responsible for such functionality, it will be in one place.

src/index.html
```diff
@@ -28,6 +28,7 @@
   <link href="https://fonts.googleapis.com/css?family=PT+Sans" rel="stylesheet">
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
   <script src="assets/plotly-latest.min.js"></script>
+  <script src="assets/string-format.js"></script>
```
gmrukwa: This should probably disappear.

PR #166, opened 2018-02-04
Prune migrated functionality

gmrukwa:
I've removed all of the code that has already been migrated to other repositories or ported otherwise.

After this merge, our repository is going to become Angular-only repo with the client itself. I've integrated Travis CI to check our builds independently on Linux. Docker image is built on each successful build and pushed onto Docker Hub.

Docker image definition is just minimalistic: copies built application onto nginx and serves it to the world.

Build process got simplified.

e2e tests are temporarily disabled, especially that no one is developing them actively.


gmrukwa:
[Prune Spectre repo to make it client-only](https://trello.com/c/IwVs3Oyw/380-prune-spectre-repo-to-make-it-client-only)

PR #168, opened 2018-03-07
wrap urls in functions

Sebaxp95:


Sebaxp95:
So joining server URL and api URL should be in apiUrl.ts?

gmrukwa:
Definitely.

PR #171, opened 2018-03-14
Create compose

gmrukwa:
Fixes the problem with tests' timeouts and introduces `docker-compose.yml` with specification of current stable services stack. In future, we will replace `client` version to use `latest` instead of `beta` or any more specific version. Right now it corresponds to actual content of our image repository.


gmrukwa:
[Dockerize main repository](https://trello.com/c/F8pJIHxP/327-dockerize-main-repository)

PR #172, opened 2018-03-14
Generic analysis form

gmrukwa:
This review introduces simple wrapper for the JSON-schema form. It will accept two URLs: for schema and layout, and exhibit `onSubmit` event in which data collected by the form will be processed.

Additionally, a bugfix for tests' results paths has been introduced.

Awaits merging of #171.


gmrukwa:
[Add generic form for scheduling analysis](https://trello.com/c/Zb9h1tPq/414-add-generic-form-for-scheduling-analysis)

src/app/generic-form/generic-form-fetch.service.spec.ts
```diff
@@ -0,0 +1,47 @@
+/*
+ * generic-form-fetch.service.spec.ts
+ * Unit tests for service fetching form definition.
+ *
+   Copyright 2017 Grzegorz Mrukwa
```
KapRom: maybe 2018? (all .ts files)

package.json
```diff
@@ -58,13 +58,15 @@
     "@angular/compiler": "^5.2.6",
     "@angular/compiler-cli": "^5.2.6",
     "@angular/core": "^5.2.6",
+    "@angular/flex-layout": "2.0.0-beta.10-4905443",
```
Sebaxp95: There is new version

gmrukwa: There is, but we've checked experimentally and this one was the newest working.

src/app/generic-form/generic-form-fetch.service.spec.ts
gmrukwa: Good catch :D

PR #173, opened 2018-03-15
Analysis view

gmrukwa:
I've added module for `analysis-view`.

URLs are preserved as they were currently on `develop`. They may be adjusted here or in #168, depending on what gets merged first.

### DIFF
--- a/src/app/analysis-view/analysis-view/analysis-view.component.html
+++ b/src/app/analysis-view/analysis-view/analysis-view.component.html
@@ -1,7 +1,23 @@
+<h1>{{algorithmName}}</h1>
+
 <mat-card *ngFor="let analysis of analyses">
     <mat-card-title>{{analysis.name}}</mat-card-title>
     <mat-card-content>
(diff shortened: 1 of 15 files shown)


gmrukwa:
[Analysis view](https://trello.com/c/2oCn8WVq/417-analysis-view)

src/app/analysis-view/analysis-view/analysis-view.component.html
```diff
@@ -0,0 +1,24 @@
...
+</mat-card>
+
+<mat-card *ngIf="errorMessage!==undefined">
+    <mat-card-title>An error occured</mat-card-title>
+    <mat-card-content>
+        {{errorMessage}}
+    </mat-card-content>
+</mat-card>
+
+<a mat-fab routerLink="../../analysis/new/{{algorithmName}}">+</a>
```
KapRom: You may try to put this in mat-action-row

gmrukwa: `mat-action-row` seems to be related only to expansion panel, so it would enforce us to hide this button. I'd leave it as it is.

PR #175, opened 2018-03-16
Table & visualization view

gmrukwa:
* collapsing of `npm list` result during the build
* simple definition of the table-view
* simple definition of a factory-alike component

PR #177, opened 2018-03-17
Aspect view

gmrukwa:
Generic aspect view with form for narrowing down the amount of data displayed and with visualization.

Will need more restyling in future, but this is out of scope of this review probably.

Waits for #175.


gmrukwa:
[aspect-view](https://trello.com/c/hMvLo73O/426-aspect-view)

PR #187, opened 2018-05-21
fix error in showing plot

Sebaxp95:


Sebaxp95:
[plotly component must remove all traces, not only the last one](https://trello.com/c/uQyAIna9/470-plotly-component-must-remove-all-traces-not-only-the-last-one)

src/app/plotly/plotly.component.ts
```diff
@@ -58,21 +58,23 @@ export class PlotlyComponent implements OnInit, AfterViewInit, OnChanges {
       console.log(this.data);
       console.log('[PlotlyComponent] layout: ');
       console.log(this.layout);
-      const div = document.getElementById('plotly-' + this.randomId);
-      Plotly.deleteTraces(div, 0);
-      Plotly.addTraces(div, this.data);
+      this.createPlot();
```
gmrukwa: Don't we need to delete the plot data somehow?

Sebaxp95: No, newPlot remove and create new plot, but in newer version of plotly is function react which is more efficient than newPlot (https://plot.ly/javascript/plotlyjs-function-reference/#plotlyreact) so I added it.

Sebaxp95: According to the documentation:
Plotly.react has the same signature as Plotly.newPlot above, and can be used in its place to **create a plot**, but when called again on the same <div> will update it far more efficiently than Plotly.newPlot

